PR #3, opened 2018-06-08
Update 'Intrinsic Functions' section

kito-cheng:


jim-wilson:
This was discussed in an issue that Andrew opened and then closed.  Andrew seems to be in favor of not documenting them.  I would prefer to document them.  It isn't possible to keep secrets like this in the free/open software world.

However, are you aware of anyone using them?  Glibc and libgcc are both using extended asms to read fp flags instead of the builtin functions.  Newlib and the linux kernel don't use them.  I'm starting to wonder why we even have them.

aswaterman:
I don't feel strongly about it.

asb:
I think the question is whether the expectation is that every RISC-V C compiler should provide these functions? If these builtins see no real use and we see no reason to actively encourage people to adopt them, then one option is to leave them undocumented here and to modify the GCC documentation to discourage their use.

palmer-dabbelt:
I think we should rename them to `riscv_internal_*` and document them as explicitly being internal to the GCC port.  How does this par if patches look?

````
diff --git a/gcc/config/riscv/riscv-builtins.c b/gcc/config/riscv/riscv-builtins.c
index b1d89fc87194..d29ef633759b 100644
--- a/gcc/config/riscv/riscv-builtins.c
+++ b/gcc/config/riscv/riscv-builtins.c
@@ -125,8 +125,8 @@ AVAIL (hard_float, TARGET_HARD_FLOAT)
   RISCV_ATYPE_##A, RISCV_ATYPE_##B
 
 static const struct riscv_builtin_description riscv_builtins[] = {
-  DIRECT_BUILTIN (frflags, RISCV_USI_FTYPE_VOID, hard_float),
-  DIRECT_NO_TARGET_BUILTIN (fsflags, RISCV_VOID_FTYPE_USI, hard_float)
+  DIRECT_BUILTIN (internal_frflags, RISCV_USI_FTYPE_VOID, hard_float),
+  DIRECT_NO_TARGET_BUILTIN (internal_fsflags, RISCV_VOID_FTYPE_USI, hard_float)
 };
 
 /* Index I is the function declaration for riscv_builtins[I], or null if the
@@ -279,8 +279,8 @@ riscv_atomic_assign_expand_fenv (tree *hold, tree *clear, tree *update)
   if (!TARGET_HARD_FLOAT)
     return;
 
-  tree frflags = GET_BUILTIN_DECL (CODE_FOR_riscv_frflags);
-  tree fsflags = GET_BUILTIN_DECL (CODE_FOR_riscv_fsflags);
+  tree frflags = GET_BUILTIN_DECL (CODE_FOR_riscv_internal_frflags);
+  tree fsflags = GET_BUILTIN_DECL (CODE_FOR_riscv_internal_fsflags);
   tree old_flags = create_tmp_var_raw (RISCV_ATYPE_USI);
 
   *hold = build2 (MODIFY_EXPR, RISCV_ATYPE_USI, old_flags,
diff --git a/gcc/config/riscv/riscv.md b/gcc/config/riscv/riscv.md
index a5940dcc4253..9c35381bb3ba 100644
--- a/gcc/config/riscv/riscv.md
+++ b/gcc/config/riscv/riscv.md
@@ -2284,13 +2284,13 @@
   ""
   "")
 
-(define_insn "riscv_frflags"
+(define_insn "riscv_internal_frflags"
   [(set (match_operand:SI 0 "register_operand" "=r")
 	(unspec_volatile [(const_int 0)] UNSPECV_FRFLAGS))]
   "TARGET_HARD_FLOAT"
   "frflags\t%0")
 
-(define_insn "riscv_fsflags"
+(define_insn "riscv_internal_fsflags"
   [(unspec_volatile [(match_operand:SI 0 "csr_operand" "rK")] UNSPECV_FSFLAGS)]
   "TARGET_HARD_FLOAT"
   "fsflags\t%0")
````

````
diff --git a/riscv-c-api.md b/riscv-c-api.md
index a40243edbaa0..38ab1fffa7cb 100644
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -56,5 +56,12 @@ https://creativecommons.org/licenses/by/4.0/.
 
 ## Intrinsic Functions
 
+### GCC-Specific Extensions
+
+#### Intrinsic Functions
+
+These functions are specific to the GCC internals and are not expected
+to be used by users.
+
+* `unsigned int __builtin_riscv_internal_frflags()`
+* `void __builtin_riscv_internal_fsflags(unsigned int flags)`
````

jim-wilson:
If someone is using these built-ins, and we change their names, then we break their code.  It isn't wise to make gcc changes like this without a good reason.  We can document them as internal without changing their names.

I'm OK with documenting them in gcc and leaving them out of this document.  And the same goes for everything else in this document.  There is no reason why every single gcc feature needs to be part of the standard API.  Palmer just started with everything that gcc supports, but we can prune it down to the things that we really need supported in all compilers if necessary.

aswaterman:
Agreed: keep names as-is, document as internal in the GCC docs, and leave out of this doc (indicating LLVM etc. are not expected to provide this API).

kito-cheng:
I've checked `sfp-machine.h` and grep all gcc source, it seems nowhere use those two intrinsic functions? just curious where they use?

kito-cheng:
I've updated those intrinsic functions to gcc specific section

cmuellner:
I believe the decision here is not to document these intrinsics in this repo.
They are currently only implemented in GCC, and there is no interest in making them available in all compilers.

GCC can decide what to do with these builtins (e.g., document them in their documentation, drop them from GCC, or simply leave the current status).

As there is nothing left to be done/decided for this repo, I'll close this PR.

PR #8, opened 2018-10-03
Add new __riscv_abi_rve macro.

jim-wilson:
Defined when -mabi=ilp32e is used.


jim-wilson:
I don't have write access, so you may as well check it in yourself.

PR #9, opened 2019-06-18
Add __riscv_vector and __riscv_bitmanip for V and B extensions.

jim-wilson:
We have assembler ports for the V and B extensions, so we should be thinking about adding preprocessor macros for them.  __riscv_vector seems straightforward.  __riscv_bitmanipulation seems inconveniently long, but __riscv_bitmanip looks OK, and is consistent with the B extension gas/gcc patches.  I asked in a riscv/riscv-bitmanip issue if they are OK with this.  We should wait for comments in case they want to suggest something different.


jim-wilson:
19 days and no comment, so I'm self approving and committing.

PR #10, opened 2020-08-11
Add __riscv_fp16 for Zfh extension

kito-cheng:
https://github.com/riscv/riscv-isa-manual/pull/496

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -54,6 +54,7 @@ https://creativecommons.org/licenses/by/4.0/.
 * `__riscv_cmodel_pic`
 * `__riscv_vector`
 * `__riscv_bitmanip`
+* `__riscv_zfh`
 
 ## Function Attributes
 


kito-cheng:
@asb @lenary @jim-wilson

lenary:
I realise this isn't like a `__riscv_float_abi_half` macro, because this is an architectural feature not an ABI feature, but maybe we should aim for `__riscv_float_isa_half` in this case? I feel we should try to match existing names somewhat rather than establishing new one-off names, though I also see how flen does not actually tell you if you have native half-width float support.

aswaterman:
@lenary I agree we need something to that effect.  To avoid ambiguity with possible future ISA support for bfloat16, maybe we should just call this macro `__riscv_zfh`?

knightsifive:
I second @aswaterman 's proposal. I'd also prefer to have `__riscv_c` (vs. `riscv_compressed`), `__riscv_b` (vs. `__riscv_bitmanip`), etc., for related macros. I don't see the benefit in adding these ad-hoc nicknames ("compressed", "bitmanip", etc.) on top of the ISA-string alphabet soup ("C", "B", etc.). If anyone else feels similarly, I would be willing to submit a PR.

kito-cheng:
`__riscv_zfh` is OK to me.

kito-cheng:
For other already existed marco `__riscv_compressed`, I would prefer keep that and do not add marco to that, but it's out of scope of this PR.

ebahapo:
Methinks that changing the pattern from adhoc to spec is worse than deviating from spec.  At least, adding the respective single letter macros in addition to the existing ones would make sense.  Anything to make life easier for the programmer.

lenary:
> Methinks that changing the pattern from adhoc to spec is worse than deviating from spec. At least, adding the respective single letter macros in addition to the existing ones would make sense. Anything to make life easier for the programmer.

For currently ratified extensions, I do agree that the ad-hoc definitions are not removable now, but we can also provide the spec-letter definitions and point to these as canonical. i.e. provide both `__riscv_compressed` (existing, ad-hoc) and `__riscv_c` (following the spec pattern).

For unratified extensions like B and V (and their `Zb*` and `Zv*` parts), I think we should only provide the letter equivalents.

PR #11, opened 2020-09-02
Reorg `Preprocessor Definitions` section.

kito-cheng:
- Tabularization, add category for those macro definitions and describe
   when will define.

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -73,7 +73,7 @@ https://creativecommons.org/licenses/by/4.0/.
 
 | Name                  | Value    | When defined                          | Alternative |
 | --------------------- | -------- | ------------------------------------- | ----------- |
-| __riscv_cmodel_pic    | 1        | GCC will defined that when program compile with `-fPIC`, `-fpic`, `-fPIE` or `-fpie`. | `__PIC__` or `__PIE__` |
+| __riscv_cmodel_pic    | 1        | GCC defines this when compiling with `-fPIC`, `-fpic`, `-fPIE` or `-fpie`. | `__PIC__` or `__PIE__` |
 
 ## Function Attributes
 
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -50,7 +50,7 @@ https://creativecommons.org/licenses/by/4.0/.
 | __riscv_bitmanip    | 1     | `B` extension is available.   |
 | __riscv_zfh         | 1     | `Zfh` extension is available. |
 
-*[1] Not all compiler provide `-mno-div` and `-mno-fdiv` option.
+*[1] Not all compilers provide `-mno-div` and `-mno-fdiv` option.
 
 ### ABI Related Preprocessor Definitions
 
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -42,14 +42,15 @@ https://creativecommons.org/licenses/by/4.0/.
 | __riscv_atomic      | 1     | `A` extension is available.   |
 | __riscv_compressed  | 1     | `C` extension is available.   |
 | __riscv_mul         | 1     | `M` extension is available.   |
-| __riscv_div         | 1     | `M` extension is available.   |
-| __riscv_muldiv      | 1     | `M` extension is available.   |
-| __riscv_fdiv        | 1     | `F` extension is available.   |
-| __riscv_fsqrt       | 1     | `F` extension is available.   |
+| __riscv_div         | 1     | `M` extension is available and `-mno-div` is not given.*[1]    |
+| __riscv_muldiv      | 1     | `M` extension is available and `-mno-div` is not given.*[1]    |
+| __riscv_fdiv        | 1     | `F` extension is available and `-mno-fdiv` is not given.*[1]   |
+| __riscv_fsqrt       | 1     | `F` extension is available and `-mno-fdiv` is not given.*[1]   |
 | __riscv_vector      | 1     | `V` extension is available.   |
 | __riscv_bitmanip    | 1     | `B` extension is available.   |
 | __riscv_zfh         | 1     | `Zfh` extension is available. |
 
+*[1] Not all compiler provide `-mno-div` and `-mno-fdiv` option.
 
 ### ABI Related Preprocessor Definitions
 
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -33,28 +33,41 @@ https://creativecommons.org/licenses/by/4.0/.
 
 ## Preprocessor Definitions
 
-* `__riscv`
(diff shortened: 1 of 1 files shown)


ebahapo:
What exactly is the PIC code model?

riscv-c-api.md
```diff
@@ -33,28 +33,41 @@ https://creativecommons.org/licenses/by/4.0/.
...
-* `__riscv_cmodel_medany`
-* `__riscv_cmodel_pic`
-* `__riscv_vector`
-* `__riscv_bitmanip`
-* `__riscv_zfh`
+| Name                | Value | When defined                  |
+| ------------------- | ----- | ----------------------------- |
+| __riscv             | 1     | Always defined.               |
+| __riscv_xlen        | <ul><li>32 for rv32</li><li>64 for rv64</li><li>128 for rv128</ul> | Always defined.             |
+| __riscv_flen        | <ul><li>128 if `Q` extension available</li><li>64 if `D` extension available</li><li>32 if `F` extension available<li></ul> | `F` extension is available. |
```
ebahapo, line 40: Or rather:

- 32 if the `F` extension is available **or**
- 64 if the `D` extension is available **or**
- 128 if the `Q` extension is available

kito-cheng:
> What exactly is the PIC code model?

In fact we treat PIC/PIE as a code model like `medlow` and `medany` in the GCC implementation, which mean if you apply -fPIC/-fpic/-fPIE/-fpie then -mcmodel is no longer matters.

This also remind me that clang don't have same behavior on this macro.

riscv-c-api.md
kito-cheng: Updated, thanks :)

jim-wilson:
The pic code model is medany with different symbol resolution, to make shared libraries work.  It is a bit of a gcc hack to call it a code model.  That perhaps made the implementation a little simpler and/or made the compiler a little easier to use.  I always considered this an internal gcc implementation detail.  I don't think that we should be exposing this outside gcc.  Though I see that we already have riscv_cmodel_pic.  Maybe we could deprecate that if no one is using it.  If you want to know if code was compiled pic, you should be using __PIC__ or __pic__ or __PIE__ or __pie__.

lenary:
I think we could just leave `__riscv_cmodel_pic` undocumented here, and potentially ensure that when in PIC-mode, gcc defines `__riscv_cmodel_medany`?

It might be good to add additional notes to `_div`, `_fdiv`, `_fsqrt` which link them to the related GCC flags. Clang unconditionally defines them all when the respective extension is enabled, because we don't have the `-mno-div` flags and the like.

ebahapo:
Ok, it seems to me then that it would be better to not add the predefined macro `__riscv_cmodel_pic` here then.  And I agree with @lenary , that PIC should then define `__riscv_cmodel_medany` as well.

Besides, PIC already defines `__PIC__`, as does PIE, which defines both the latter and `__PIE__`.

kito-cheng:
ChangeLog
- Move `__riscv_cmodel_pic` to `Deprecated Preprocessor Definitions`.
- Mention `__riscv_div`, `__riscv_muldiv`, `__riscv_fdiv` and `__riscv_fsqrt` won't defined when `-mno-div` and `-mno-fdiv` option is given

riscv-c-api.md
```diff
@@ -33,28 +33,47 @@ https://creativecommons.org/licenses/by/4.0/.
...
+| Name                  | Value    | When defined                          |
+| --------------------- | -------- | ------------------------------------- |
+| __riscv_cmodel_medlow | 1        | Defined if using `medlow` code model. |
+| __riscv_cmodel_medany | 1        | Defined if using `medany` code model. |
+
+### Deprecated Preprocessor Definitions
+
+| Name                  | Value    | When defined                          | Alternative |
+| --------------------- | -------- | ------------------------------------- | ----------- |
+| __riscv_cmodel_pic    | 1        | GCC will defined that when program compile with `-fPIC`, `-fpic`, `-fPIE` or `-fpie`. | `__PIC__` or `__PIE__` |
```
lenary, line 76:
```suggestion
| __riscv_cmodel_pic    | 1        | GCC defines this when compiling with `-fPIC`, `-fpic`, `-fPIE` or `-fpie`. | `__PIC__` or `__PIE__` |
```

lenary (approved):
One grammar nit, then LGTM.

riscv-c-api.md
kito-cheng: Cool! this is the first time I have seen this github feature :P

kito-cheng:
@jim-wilson ping

riscv-c-api.md
```diff
@@ -33,28 +33,47 @@ https://creativecommons.org/licenses/by/4.0/.
...
+| __riscv_mul         | 1     | `M` extension is available.   |
+| __riscv_div         | 1     | `M` extension is available and `-mno-div` is not given.*[1]    |
+| __riscv_muldiv      | 1     | `M` extension is available and `-mno-div` is not given.*[1]    |
+| __riscv_fdiv        | 1     | `F` extension is available and `-mno-fdiv` is not given.*[1]   |
+| __riscv_fsqrt       | 1     | `F` extension is available and `-mno-fdiv` is not given.*[1]   |
+| __riscv_vector      | 1     | `V` extension is available.   |
+| __riscv_bitmanip    | 1     | `B` extension is available.   |
+| __riscv_zfh         | 1     | `Zfh` extension is available. |
+
+*[1] Not all compiler provide `-mno-div` and `-mno-fdiv` option.
```
jim-wilson, line 53:
```suggestion
*[1] Not all compilers provide `-mno-div` and `-mno-fdiv` option.
```

jim-wilson (approved):
Looks good with one minor English improvement.

This will require a gcc patch to define __riscv_cmodel_medany when PIC.  Are you writing that patch or am I?

kito-cheng:
@jim-wilson 
> This will require a gcc patch to define __riscv_cmodel_medany when PIC. Are you writing that patch or am I?
I am writing that, I'll send that today :)

vineetgarc:
> > What exactly is the PIC code model?
> 
> In fact we treat PIC/PIE as a code model like `medlow` and `medany` in the GCC implementation, which mean if you apply -fPIC/-fpic/-fPIE/-fpie then -mcmodel is no longer matters.

@kito-cheng Is this still true ?

kito-cheng:
>> > What exactly is the PIC code model?
>>
>> In fact we treat PIC/PIE as a code model like medlow and medany in the GCC implementation, which mean if you apply -fPIC/-fpic/-fPIE/-fpie then -mcmodel is no longer matters.
>
>@kito-cheng Is this still true ?

We didn't change those part for a long time, so, yes.

PR #13, opened 2020-11-09
Introduce new scheme of extension related predefined macro

kito-cheng:
The motivation of this scheme is have an unify naming scheme for
extension macro and add the capability to checking version.

The new naming scheme is `__riscv_<extension name>`, and it will define
the value to `<MAJOR_VERSION> * 1,000,000 + <MINOR_VERSION> * 1,000 + <REVISION_VERSION>`

For example:
 - F-extension v2.2 will define `__riscv_f` as `20002000`.
 - B-extension v0.92 will define `__riscv_b` as `92000`.

We also introduce a new macro `__riscv_arch_test_marco` to check this
test marco set is available.

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -55,8 +55,8 @@ https://creativecommons.org/licenses/by/4.0/.
 
 Architecture extension test macro is a new set of test macro to checking the
 availability and version for certain extension, however not all compilers are
-supported, so you should check `__riscv_arch_test_macro` to make sure this
-compiler is supporting those preprocessor definitions.
+supported, so you should check `__riscv_arch_test` to make sure this compiler
+is supporting those preprocessor definitions.
 
 The value of architecture extension test macro are defined as its version,
 which is compute by following formula:
@@ -71,7 +71,7 @@ For example:
 
 | Name                    | Value        | When defined                  |
 | ----------------------- | ------------ | ----------------------------- |
-| __riscv_arch_test_macro | 1            | Defined if compiler support new architecture extension test macro. |
+| __riscv_arch_test       | 1            | Defined if compiler support new architecture extension test macro. |
 | __riscv_i               | Arch Version | `I` extension is available.   |
 | __riscv_e               | Arch Version | `E` extension is available.   |
 | __riscv_m               | Arch Version | `M` extension is available.   |
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -39,15 +39,11 @@ https://creativecommons.org/licenses/by/4.0/.
 | __riscv_xlen        | <ul><li>32 for rv32</li><li>64 for rv64</li><li>128 for rv128</ul> | Always defined.             |
 | __riscv_flen        | <ul><li>32 if the F extension is available **or**</li><li>64 if `D` extension available **or**</li><li>128 if `Q` extension available</li></ul> | `F` extension is available. |
 | __riscv_32e         | 1     | `E` extension is available.   |
-| __riscv_atomic      | 1     | `A` extension is available.   |
-| __riscv_compressed  | 1     | `C` extension is available.   |
 | __riscv_mul         | 1     | `M` extension is available.   |
 | __riscv_div         | 1     | `M` extension is available and `-mno-div` is not given.*[1]    |
 | __riscv_muldiv      | 1     | `M` extension is available and `-mno-div` is not given.*[1]    |
 | __riscv_fdiv        | 1     | `F` extension is available and `-mno-fdiv` is not given.*[1]   |
 | __riscv_fsqrt       | 1     | `F` extension is available and `-mno-fdiv` is not given.*[1]   |
-| __riscv_vector      | 1     | `V` extension is available.   |
-| __riscv_bitmanip    | 1     | `B` extension is available.   |
 
 *[1] Not all compilers provide `-mno-div` and `-mno-fdiv` option.
 
@@ -105,6 +101,11 @@ For example:
 | Name                  | Value    | When defined                          | Alternative |
 | --------------------- | -------- | ------------------------------------- | ----------- |
 | __riscv_cmodel_pic    | 1        | GCC defines this when compiling with `-fPIC`, `-fpic`, `-fPIE` or `-fpie`. | `__PIC__` or `__PIE__` |
+| __riscv_atomic        | 1        | `A` extension is available.   |
+| __riscv_compressed    | 1        | `C` extension is available.   |
+| __riscv_vector        | 1        | `V` extension is available.   |
+| __riscv_bitmanip      | 1        | `B` extension is available.   |
+
 
 ## Function Attributes
 
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -48,9 +48,40 @@ https://creativecommons.org/licenses/by/4.0/.
 | __riscv_fsqrt       | 1     | `F` extension is available and `-mno-fdiv` is not given.*[1]   |
 | __riscv_vector      | 1     | `V` extension is available.   |
 | __riscv_bitmanip    | 1     | `B` extension is available.   |
-| __riscv_zfh         | 1     | `Zfh` extension is available. |
 
 *[1] Not all compilers provide `-mno-div` and `-mno-fdiv` option.
+
+### Architecture Extension Test Macro
+
+Architecture extension test macro is a new set of test macro to checking the
+availability and version for certain extension, however not all compilers are
+supported, so you should check `__riscv_arch_test_macro` to make sure this
+compiler is supporting those preprocessor definitions.
+
+The value of architecture extension test macro are defined as its version,
+which is compute by following formula:
+
+```
+<MAJOR_VERSION> * 1,000,000 + <MINOR_VERSION> * 1,000 + <REVISION_VERSION>
+```
+
+For example:
+- F-extension v2.2 will define `__riscv_f` as `20002000`.
+- B-extension v0.92 will define `__riscv_b` as `92000`.
+
+| Name                    | Value        | When defined                  |
+| ----------------------- | ------------ | ----------------------------- |
+| __riscv_arch_test_macro | 1            | Defined if compiler support new architecture extension test macro. |
+| __riscv_i               | Arch Version | `I` extension is available.   |
+| __riscv_e               | Arch Version | `E` extension is available.   |
+| __riscv_m               | Arch Version | `M` extension is available.   |
+| __riscv_a               | Arch Version | `A` extension is available.   |
+| __riscv_f               | Arch Version | `F` extension is available.   |
+| __riscv_d               | Arch Version | `D` extension is available.   |
+| __riscv_c               | Arch Version | `C` extension is available.   |
+| __riscv_b               | Arch Version | `B` extension is available.   |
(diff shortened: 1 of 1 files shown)


riscv-c-api.md
```diff
@@ -48,10 +48,41 @@ https://creativecommons.org/licenses/by/4.0/.
 | __riscv_fsqrt       | 1     | `F` extension is available and `-mno-fdiv` is not given.*[1]   |
 | __riscv_vector      | 1     | `V` extension is available.   |
 | __riscv_bitmanip    | 1     | `B` extension is available.   |
-| __riscv_zfh         | 1     | `Zfh` extension is available. |
 
 *[1] Not all compilers provide `-mno-div` and `-mno-fdiv` option.
 
+### Architecture Extension Test Marco
```
jim-wilson, line 54: Marco -> Macro
this happens in a number of places

riscv-c-api.md
```diff
@@ -48,10 +48,41 @@ https://creativecommons.org/licenses/by/4.0/.
...
+
+The value of architecture extension test marco are defined as its version,
+which is compute by following formula:
+
+```
+<MAJOR_VERSION> * 1,000,000 + <MINOR_VERSION> * 1,000 + <REVISION_VERSION>
+```
+
+For example:
+- F-extension v2.2 will define `__riscv_f` as `2002000`.
```
jim-wilson, line 69: Looks like you are missing a zero here.

jim-wilson (approved):
Looks reasonable to me with a few typos fixed.

riscv-c-api.md
```diff
@@ -48,10 +48,41 @@ https://creativecommons.org/licenses/by/4.0/.
...
+- B-extension v0.92 will define `__riscv_b` as `92000`.
+
+| Name                    | Value        | When defined                  |
+| ----------------------- | ------------ | ----------------------------- |
+| __riscv_arch_test_marco | 1            | Defined if compiler support new architecture extension test marco. |
```
ebahapo, line 74: It's better to move this comment before the table.  He it looks like a macro itself.

riscv-c-api.md
kito-cheng: Fixed

riscv-c-api.md
kito-cheng: Fixed

riscv-c-api.md
kito-cheng: Not sure you mean which comment? do you mind provide diff for that?

riscv-c-api.md
```diff
@@ -48,10 +48,41 @@ https://creativecommons.org/licenses/by/4.0/.
...
+which is compute by following formula:
+
+```
+<MAJOR_VERSION> * 1,000,000 + <MINOR_VERSION> * 1,000 + <REVISION_VERSION>
+```
+
+For example:
+- F-extension v2.2 will define `__riscv_f` as `20002000`.
+- B-extension v0.92 will define `__riscv_b` as `92000`.
+
```
ebahapo, line 71: Add "The macros below are defined to `1` if the respective architecture extension is supported:"

riscv-c-api.md
```diff
@@ -48,10 +48,41 @@ https://creativecommons.org/licenses/by/4.0/.
...
+| Name                    | Value        | When defined                  |
+| ----------------------- | ------------ | ----------------------------- |
+| __riscv_arch_test_macro | 1            | Defined if compiler support new architecture extension test macro. |
```
ebahapo, line 74: Delete this line.

kito-cheng: It's a macro to indicate this compiler has supported those macro. 
e.g.
```
#ifdef __riscv_arch_test_macro
// We can assume __riscv_i and __riscv_<ext> is supported on this compiler.
#endif
```

kito-cheng: The intention of this macro is I believe there would be an junction period, some compiler provide those macro, and some compiler not ready yet, and this macro can tell user it's supported or not.

ebahapo: Methinks that it's not necessary.  The respective macros can be tested for `#ifdef`.

ebahapo:
On this topic, just for the record, there already is a legacy macro to test for the M extension.  It may be too late to immediately deprecate it in the compilers, as discussed in the conference call, but not to do so in the spec, favoring `__riscv_m` instead.

kito-cheng:
> On this topic, just for the record, there already is a legacy macro to test for the M extension. It may be too late to immediately deprecate it in the compilers, as discussed in the conference call, but not to do so in the spec, favoring __riscv_m instead.

I agree with that, we should deprecating `__riscv_mul`, `__riscv_div`, `__riscv_muldiv`, `__riscv_fdiv`, and `__riscv_fsqrt`, but I would like push the timing little bit later, since there is no open source compilers has supported new macro in there latest formal release yet, so I would suggest deprecated that after GCC 11 and LLVM 12 release.

riscv-c-api.md
kito-cheng: I would prefer keep this macro, if user only want to use new macro, they can have way to checking compiler is supported, and maybe emit error when compiler not supported, without the macro we need to tested that by `#if defined(__riscv_i) || defined(__riscv_e)`.

```
#ifndef __riscv_arch_test_macro
#error Compiler too old, plz upgrade your RISC-V compiler.
#endif
```

lenary:
> > On this topic, just for the record, there already is a legacy macro to test for the M extension. It may be too late to immediately deprecate it in the compilers, as discussed in the conference call, but not to do so in the spec, favoring __riscv_m instead.
> 
> I agree with that, we should deprecating `__riscv_mul`, `__riscv_div`, `__riscv_muldiv`, `__riscv_fdiv`, and `__riscv_fsqrt`, but I would like push the timing little bit later, since there is no open source compilers has supported new macro in there latest formal release yet, so I would suggest deprecated that after GCC 11 and LLVM 12 release.

I'm in agreement here. I think we can deprecate the old macros in this document immediately, and compilers can make their own choices about how soon they deprecate their definitions, and how long they want to maintain compatibility for.

riscv-c-api.md
ebahapo: Fair enough.  However, a macro doesn't have to be named `..._macro`.

jim-wilson:
__riscv_muldiv maps directly to the M extension and is redundant.

But the others, mul, div, fdiv, fsqrt do not directly map to any official extension.  They do map to gcc options, but maybe not llvm options.  People in the zfinx task group are particularly interested in some of these options, because dropping certain instructions like div or fdiv might allow smaller cores.  Also, academic projects sometimes drop div and/or fdiv to save time as it is more difficult for a student to implement.  Of course, maybe they don't have to be official macros and we just leave them as gcc extensions but that could be inconvenient if llvm doesn't implement them.  I believe there are some people using them.

Note that Kito's recent newlib change to add a sqrt stub function uses __riscv_fsqrt.  longlong.h (originally GMP but now everywhere) uses __riscv_ mul.  Those are the only ones I see an obvious use for in the GNU toolchain.  I didn't check other software.

lenary:
LLVM defines `__riscv_mul`, `__riscv_div` and `__riscv_muldiv` together if you ask for the M extension, and does not allow you to enable M partially. The same is true for F/D, where if you have `__riscv_flen != 0`, then you also have `__riscv_fdiv` and `__riscv_fsqrt`. 

I'm happy to take the point that `__riscv_m` would be the new name for `__riscv_muldiv`, not `__riscv_mul`. 

I think the partial support macros (`__riscv_mul`, `__riscv_div`, `__riscv_fdiv` and `__riscv_fsqrt`) should probably be deprecated, or at least made more informal - LLVM will define them when they're supported, but I'm not sure we're interested in `-fno-div` etc - I don't know of anyone who has specifically requested it.

jim-wilson:
Removing div has come up in various discussions.  For instance, in the zfinx spec proposal
    https://github.com/riscv/riscv-zfinx/blob/master/Zfinx_spec.adoc
under rationale 2 iv it mentions optionally removing fdiv to save space.  And in the last section FP Config to Reduce Area, it mentions the gcc -fno-fdiv option, and proposes a Zfdiv to indicate whether divide is implemented or not, where Zfbase does not include Zfdiv.  They also want to make multiply-add optional, but we don't have existing support for that.  This stuff would only be used in small low cost embedded chips.

kito-cheng:
Changes:
- Rename __riscv_arch_test_macro to __riscv_arch_test
- Mark __riscv_atomic, __riscv_compressed, __riscv_vector and __riscv_bitmanip as deprecated.

riscv-c-api.md
kito-cheng: Fixed

kito-cheng:
Changes:
- Mark __riscv_mul, __riscv_div, __riscv_muldiv, __riscv_fdiv and __riscv_fsqrt as deprecated.

kito-cheng:
I guess eventually we should have bunch of Z-extension to represent such combination: w/ or w/o for fdiv, fsqrt, div, then we have formal way to describe that instead of a legacy feature on GCC, but it seems really terrible # of combination and fragmentation to me.

kito-cheng:
@lenary @ebahapo ping, do you mind give LGTM if current version is OK for you?

lenary:
> I guess eventually we should have bunch of Z-extension to represent such combination: w/ or w/o for fdiv, fsqrt, div, then we have formal way to describe that instead of a legacy feature on GCC, but it seems really terrible # of combination and fragmentation to me.

the ABI seems unable to stand in the way of the fragmentation - but I think if we say "yes, when there's `Zfdiv`, then we provide `__riscv_zfdiv`", then that fits this proposed scheme, without having to bend over backwards today.

riscv-c-api.md
```diff
@@ -74,6 +94,17 @@ https://creativecommons.org/licenses/by/4.0/.
...
 | __riscv_cmodel_pic    | 1        | GCC defines this when compiling with `-fPIC`, `-fpic`, `-fPIE` or `-fpie`. | `__PIC__` or `__PIE__` |
+| __riscv_mul           | 1        | `M` extension is available.   |
+| __riscv_div           | 1        | `M` extension is available and `-mno-div` is not given.*[1]    |
+| __riscv_muldiv        | 1        | `M` extension is available and `-mno-div` is not given.*[1]    |
+| __riscv_atomic        | 1        | `A` extension is available.   |
+| __riscv_fdiv          | 1        | `F` extension is available and `-mno-fdiv` is not given.*[1]   |
+| __riscv_fsqrt         | 1        | `F` extension is available and `-mno-fdiv` is not given.*[1]   |
+| __riscv_compressed    | 1        | `C` extension is available.   |
+| __riscv_vector        | 1        | `V` extension is available.   |
+| __riscv_bitmanip      | 1        | `B` extension is available.   |
```
lenary, line 105: Can you document the Alternatives?

kito-cheng:
Changes:
- Add `alternative` for  `__riscv_atomic`, `__riscv_compressed`, `__riscv_vector`, `__riscv_bitmanip`, `__riscv_mul`, `__riscv_div`, `__riscv_muldiv`, `__riscv_fdiv` and `__riscv_fsqrt`

riscv-c-api.md
kito-cheng: Good suggestion, updated.

lenary (approved):
LGTM!

PR #14, opened 2020-12-09
Add riscv_int.h and define int_xlen_t and uint_xlen_t

kito-cheng:
This PR is the product of this [RFC](https://github.com/riscv/riscv-elf-psabi-doc/issues/158)[1], the main purpose is provide a pair of integer type with XLEN-bits.

According the discussion, many people are concern about we should not add the type directly, that will corrupt the global namespace, so it would be great if we put it in the header files.

So why it put in `riscv_intrinsic.h`? here is two reason, first reason is the initial request are come from intrinsic function interface, so put they together should be reasonable, second reason is I have no idea which file name is the best :P `riscv.h` is little bit too generic and sounds like easy to conflict with exist headers - as I know at least gcc, binutils and gdb has one.

[1] https://github.com/riscv/riscv-elf-psabi-doc/issues/158

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -238,7 +238,7 @@ vint8m1_t __riscv_vadd_vv_i8m1(vint8m1_t vs2, vint8m1_t vs1, size_t vl); // vadd
 ### riscv_int.h
 
 This header file defined several common RISC-V specific integer type and related
-marco could be used in intrinsic interface or user program directly.
+macro could be used in intrinsic interface or user program directly.
 
 | Type Name             | Meaning                              |
 | --------------------- | ------------------------------------ |
@@ -250,17 +250,17 @@ marco could be used in intrinsic interface or user program directly.
 | __RISCV_INT_XLEN_MAX  | Maximum value of int_xlen_t          |
 | __RISCV_INT_XLEN_MIN  | Minimum value of int_xlen_t          |
 | __RISCV_UINT_XLEN_MAX | Maximum value of uint_xlen_t         |
-| __RISCV_PRIdXLEN      | Expand to a string literal for use as a `d` print specificier for int_xlen_t/uint_xlen_t. |
-| __RISCV_PRIiXLEN      | Expand to a string literal for use as a `i` print specificier for int_xlen_t/uint_xlen_t. |
-| __RISCV_PRIoXLEN      | Expand to a string literal for use as a `o` print specificier for int_xlen_t/uint_xlen_t. |
-| __RISCV_PRIuXLEN      | Expand to a string literal for use as a `u` print specificier for int_xlen_t/uint_xlen_t. |
-| __RISCV_PRIxXLEN      | Expand to a string literal for use as a `x` print specificier for int_xlen_t/uint_xlen_t. |
-| __RISCV_PRIXXLEN      | Expand to a string literal for use as a `X` print specificier for int_xlen_t/uint_xlen_t. |
-| __RISCV_SCNdXLEN      | Expand to a string literal for use as a `d` scan specificier for int_xlen_t/uint_xlen_t. |
-| __RISCV_SCNiXLEN      | Expand to a string literal for use as a `i` scan specificier for int_xlen_t/uint_xlen_t. |
-| __RISCV_SCNoXLEN      | Expand to a string literal for use as a `o` scan specificier for int_xlen_t/uint_xlen_t. |
-| __RISCV_SCNuXLEN      | Expand to a string literal for use as a `u` scan specificier for int_xlen_t/uint_xlen_t. |
-| __RISCV_SCNxXLEN      | Expand to a string literal for use as a `x` scan specificier for int_xlen_t/uint_xlen_t. |
+| __RISCV_PRIdXLEN      | Expand to a string literal for use as a `d` print specifier for int_xlen_t/uint_xlen_t. |
+| __RISCV_PRIiXLEN      | Expand to a string literal for use as a `i` print specifier for int_xlen_t/uint_xlen_t. |
+| __RISCV_PRIoXLEN      | Expand to a string literal for use as a `o` print specifier for int_xlen_t/uint_xlen_t. |
+| __RISCV_PRIuXLEN      | Expand to a string literal for use as a `u` print specifier for int_xlen_t/uint_xlen_t. |
+| __RISCV_PRIxXLEN      | Expand to a string literal for use as a `x` print specifier for int_xlen_t/uint_xlen_t. |
+| __RISCV_PRIXXLEN      | Expand to a string literal for use as a `X` print specifier for int_xlen_t/uint_xlen_t. |
+| __RISCV_SCNdXLEN      | Expand to a string literal for use as a `d` scan specifier for int_xlen_t/uint_xlen_t. |
+| __RISCV_SCNiXLEN      | Expand to a string literal for use as a `i` scan specifier for int_xlen_t/uint_xlen_t. |
+| __RISCV_SCNoXLEN      | Expand to a string literal for use as a `o` scan specifier for int_xlen_t/uint_xlen_t. |
+| __RISCV_SCNuXLEN      | Expand to a string literal for use as a `u` scan specifier for int_xlen_t/uint_xlen_t. |
+| __RISCV_SCNxXLEN      | Expand to a string literal for use as a `x` scan specifier for int_xlen_t/uint_xlen_t. |
 
 
 ## Constraints on Operands of Inline Assembly Statements
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -235,6 +235,27 @@ long __riscv_clmul (long a, long b); // clmul rd, rs1, rs2
 vint8m1_t __riscv_vadd_vv_i8m1(vint8m1_t vs2, vint8m1_t vs1, size_t vl); // vadd.vv vd, vs2, vs1
 ```
 
+### riscv_intrinsic.h
+
+This file is universal intrinsic header file for all extensions, each extension
+could have their own header files, but should included in this file if
+corresponding extension is enabled.
+
+This header file also defined several common RISC-V specific type and related
+marco could be used in intrinsic interface or user program directly.
+
+| Type Name             | Meaning                              |
+| --------------------- | ------------------------------------ |
+| int_xlen_t            | Signed integer type with XLEN bits   |
+| uint_xlen_t           | Unsigned integer type with XLEN bits |
+
+| Macro Name            | Value                                |
+| --------------------- | ------------------------------------ |
+| INT_XLEN_MAX          | Minimum value of int_xlen_t          |
+| INT_XLEN_MIN          | Minimum value of int_xlen_t          |
+| UINT_XLEN_MAX         | Maximum value of uint_xlen_t         |
+| UINT_XLEN_MIN         | Minimum value of uint_xlen_t         |
+
 ## Constraints on Operands of Inline Assembly Statements
 
 This section lists operand constraints that can be used with inline assembly
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -235,26 +235,33 @@ long __riscv_clmul (long a, long b); // clmul rd, rs1, rs2
 vint8m1_t __riscv_vadd_vv_i8m1(vint8m1_t vs2, vint8m1_t vs1, size_t vl); // vadd.vv vd, vs2, vs1
 ```
 
-### riscv_intrinsic.h
+### riscv_int.h
 
-This file is universal intrinsic header file for all extensions, each extension
-could have their own header files, but should included in this file if
-corresponding extension is enabled.
-
-This header file also defined several common RISC-V specific type and related
+This header file defined several common RISC-V specific integer type and related
 marco could be used in intrinsic interface or user program directly.
 
 | Type Name             | Meaning                              |
 | --------------------- | ------------------------------------ |
-| int_xlen_t            | Two's-complement signed integer type with exactly XLEN bits (no padding bits)   |
-| uint_xlen_t           | Unsigned integer type with exactly XLEN bits (no padding bits) |
+| __riscv_int_xlen_t    | Two's-complement signed integer type with exactly XLEN bits (no padding bits)   |
+| __riscv_uint_xlen_t   | Unsigned integer type with exactly XLEN bits (no padding bits) |
 
 | Macro Name            | Value                                |
 | --------------------- | ------------------------------------ |
-| INT_XLEN_MAX          | Maximum value of int_xlen_t          |
-| INT_XLEN_MIN          | Minimum value of int_xlen_t          |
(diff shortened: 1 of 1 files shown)


jim-wilson:
The B extension proposal calls for a rvintrin.h file.
https://github.com/riscv/riscv-bitmanip/blob/master/texsrc/bext.tex#L2403
https://github.com/riscv/riscv-bitmanip/blob/master/cproofs/rvintrin.h
I don't care about the file name.  Either rvintrin.h or riscv_intrinsic.h is OK with me.  Or anything else is reasonable.  Looking at gcc ports, I see a lot more examples of the *intrin.h pattern than the *intrinsic.h pattern.

The B extension is proposing that there is that we have a single intrinsic header file for riscv, though given the size of the V extension, I'm not sure if it is a good idea to force people to include the V support when they don't want it.  Maybe put each extension in its own file, and then include it in the master (rvintrin.h?) file, maybe only if the appropriate extension is enabled.

I don't care about the type names.  People have asked for this feature before, so adding the types is useful.

It isn't obvious what the MIN MAX macros are for, or how the types can hvae min/max sizes.

knightsifive:
@jim-wilson I think the macros are meant to be the natural extensions of ones in `limits.h` and `stdint.h`.

I wonder if it's worth specifying that `int_xlen_t` is two's complement, and both `int_xlen_t` and `uint_xlen_t` have no padding bits (cf. `stdint.h`).

riscv-c-api.md
```diff
@@ -114,7 +114,25 @@ For example:
...
+This file is universal intrinsic header file for all extensions, each extension
+could have their own header files, but should included in this file if
+corresponding extension is enabled.
+
+This header file also defined several common RISC-V specific type and related
+marco could be used in intrinsic interface or user program directly.
+
+| Type Name             | Meaning                              |
+| --------------------- | ------------------------------------ |
+| int_xlen_t            | Signed integer type with XLEN bits   |
```
knightsifive, line 130:
```suggestion
| int_xlen_t            | 2's complement signed integer type with XLEN bits (no padding bits)   |
```

riscv-c-api.md
```diff
@@ -114,7 +114,25 @@ For example:
...
+| uint_xlen_t           | Unsigned integer type with XLEN bits |
```
knightsifive, line 131:
```suggestion
| uint_xlen_t           | Unsigned integer type with exactly XLEN bits (no padding bits) |
```

riscv-c-api.md
```diff
@@ -114,7 +114,25 @@ For example:
...
+
+| Macro Name            | Value                                |
+| --------------------- | ------------------------------------ |
+| INT_XLEN_MAX          | Minimum value of int_xlen_t          |
```
knightsifive, line 135:
```suggestion
| INT_XLEN_MAX          | Maximum value of int_xlen_t          |
```

jim-wilson:
My brain was stuck on type sizes, which don't have min and mix, but min and max representable values makes perfect sense.

knightsifive:
While we're at it, there's also the `U?INT_XLEN_C`, `PRI[diuoxX]XLEN`, and `SCN[diuox[XLEN` macros.
https://en.cppreference.com/w/c/types/integer
Happy to add these in a separate PR.

kito-cheng:
> The B extension is proposing that there is that we have a single intrinsic header file for riscv, though given the size of the V extension, I'm not sure if it is a good idea to force people to include the V support when they don't want it. Maybe put each extension in its own file, and then include it in the master (rvintrin.h?) file, maybe only if the appropriate extension is enabled.

Oh, maybe my word not clear enough, but I intend to do that, e.g. only included V-ext intrinsic when V or zv* extensions is enabled.

```
+ This file is universal intrinsic header file for all extensions, each extension
+ could have their own header files, but should included in this file if
+ corresponding extension is enabled.
```

kito-cheng:
> While we're at it, there's also the `U?INT_XLEN_C`, `PRI[diuoxX]XLEN`, and `SCN[diuox]XLEN` macros.
> https://en.cppreference.com/w/c/types/integer
> Happy to add these in a separate PR.

@knightsifive thanks, in fact I was refer to the same page before :P

nick-knight (approved):
Looks good to me.

cmuellner:
I think there are two issues to resolve/decide to land this:

* gating the types behind a header (and if so, which header file name)
* chose an alternative name to avoid namespace collision or at least namespace pollution

I want to avoid these types being gated by a header file with the name "intrinsic" in it. Intrinsics are just some use cases. So I am for either nothing or `__riscv.h` (or similar).

If namespace pollution is a concern, then I am also ok with a name like `__riscv_int_xlen_t` and `__riscv_uint_xlen_t`.

I also agree about creating a complete set of support macros. However, we should drop `UINT_XLEN_MIN` from the PR.

kito-cheng:
I just narrow down the scope of this PR, and made this specific to adding xlen-integer type only, also add print/scan modifier macros.

Also rename the file to riscv_int.h and adding prefix `__RISCV_`/`__riscv_`.

kito-cheng:
cc. @asb @topperc @JeffreyALaw @palmer-dabbelt @preames

cmuellner:
@kito-cheng, do you have a particular use case for these types?

Last time we discussed this (I raised the topic in the Toolchain SIG call), we only identified shared code for RV32/RV64 that accesses CSRs as a potential candidate. But that could be solved differently as well and would not really justify the introduction of a new type. Therefore, we deferred this topic back then.

kito-cheng:
@cmuellner I guess I didn't give the context why I restart that again, the reason I restart is because @topperc is working on scalar intrinsic like #44, we discuss some of intrinsic other than scalar crypto might need xlen type, but `long` is not really suitable for that since [ilp32/rv64](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/381) seems not so far (okay, still long road, but not far as before :P), so I think it good timing to restart and also including your suggestion before (I was really dislike `__riscv_` prefix but those potential issue is more important than personal preference/taste).

LLVM implementation: https://reviews.llvm.org/D154706

riscv-c-api.md
```diff
@@ -235,6 +235,34 @@ long __riscv_clmul (long a, long b); // clmul rd, rs1, rs2
...
+| --------------------- | ------------------------------------ |
+| __riscv_int_xlen_t    | Two's-complement signed integer type with exactly XLEN bits (no padding bits)   |
+| __riscv_uint_xlen_t   | Unsigned integer type with exactly XLEN bits (no padding bits) |
+
+| Macro Name            | Value                                |
+| --------------------- | ------------------------------------ |
+| __RISCV_INT_XLEN_MAX  | Maximum value of int_xlen_t          |
+| __RISCV_INT_XLEN_MIN  | Minimum value of int_xlen_t          |
+| __RISCV_UINT_XLEN_MAX | Maximum value of uint_xlen_t         |
+| __RISCV_PRIdXLEN      | Expand to a string literal for use as a `d` print specificier for int_xlen_t/uint_xlen_t. |
```
topperc, line 253: specifier*

riscv-c-api.md
```diff
@@ -235,6 +235,34 @@ long __riscv_clmul (long a, long b); // clmul rd, rs1, rs2
 vint8m1_t __riscv_vadd_vv_i8m1(vint8m1_t vs2, vint8m1_t vs1, size_t vl); // vadd.vv vd, vs2, vs1
 ```
 
+### riscv_int.h
+
+This header file defined several common RISC-V specific integer type and related
+marco could be used in intrinsic interface or user program directly.
```
topperc, line 241: macro*

kito-cheng:
Changes:

- Fix typos

cmuellner (approved):

topperc:
I'm not convinced of the need for this. I don't see a strong use case of writing xlen aware bitmanipulation code. Some crypto algorithms need to be different for RV32 and RV64, but once you've done that you know what xlen is and can use the 32/64 bitmanip intrinsics directly as needed.

Timmmm:
This would be useful for writing very low level code that interacts with inline assembly. E.g. to read CSRs.

Timmmm:
> we only identified shared code for RV32/RV64 that accesses CSRs as a potential candidate. But that could be solved differently

Ha funny that you identified the exact use case that led me to this PR. Can you expand on how it can be solved differently (without just writing this file myself, which is obviously what I've done for now).

cmuellner:
> > we only identified shared code for RV32/RV64 that accesses CSRs as a potential candidate. But that could be solved differently
> 
> Ha funny that you identified the exact use case that led me to this PR. Can you expand on how it can be solved differently (without just writing this file myself, which is obviously what I've done for now).

Then you found it!

We discussed a similar proposal yesterday: the introduction of pseudo instructions for xlen-wide loads/stores.
While there are use cases that benefit from abstracting RV32/RV64 differences, these use cases are rare.
Therefore, the costs of introducing a data type and all related support exceed the benefits.

cmuellner:
Closing as not planned.

Timmmm:
> Then you found it!
> Therefore, the costs of introducing a data type and all related support exceed the benefits.

I am unconvinced. The costs seem insignificant. Well if anyone is reading this, it seems like fortunately most people have chosen [the same name](https://grep.app/search?q=int_xlen_t).

preames:
> We discussed a similar proposal yesterday: the introduction of pseudo instructions for xlen-wide loads/stores.
> While there are use cases that benefit from abstracting RV32/RV64 differences, these use cases are rare.
> Therefore, the costs of introducing a data type and all related support exceed the benefits.

I think we may have accidentally commingled two things here.  I had made the point in the toolchain SIG that the community of folks hand writing assembly that needs to run on both rv32 and rv64 is likely quite small, and that adding complexity to all tool-chain implementations would be hard to justify under that assumption.  (To be clear here - "hard" does not mean "impossible"; it just means significant justification is needed.)

My point (which is what I think you're referencing here) was *not* in relation to whether we wanted/needed support for C/C++ language source which is compiled to both rv32 and rv64.  That's a different, possibly much larger, user community.  To be fair, the discussion on this issue to date doesn't seem to make that distinction itself very well - as the major use case mentioned is in fact interacting with inline assembly.  

To be clear, I am not speaking in support of this proposal; I currently have no opinion on its merits.  I'm only noting that the justification required for a C/C++ feature and an assembler feature are _different_ and we should be careful not to commingle them,  

If this feature does have a motivating use which is not interacting with inline assembly, it would be good if the advocates would explicitly spell it out.

cmuellner:
> I am unconvinced. The costs seem insignificant. Well if anyone is reading this, it seems like fortunately most people have chosen [the same name](https://grep.app/search?q=int_xlen_t).

I don't know if grep.app found all instances on GitHub, but let's go through all of the found projects and look into the use cases:

1. `riscv/riscv-crypto`
   This is an initial proposal for scalar crypto intrinsics.
   Meanwhile, we have merged scalar crypto instrinsics without `[u]int_xlen_t`.

2. `Nuclei-Software/nuclei-board-labs`
   Again, different scalar crypto intrinsics.

3. `chipsalliance/Cores-VeeR-EH2`
   Use of uint_xlen_t in comments (explaining an instruction).

4. `five-embeddev/riscv-scratchpad` and `five-embeddev/build-and-verify`
   CSR access

5. `nakane1chome/modern-cxx-riscv`
   CSR access

6. `uw-unsat/serval`
   Fallback implementation of `bext` instruction.

7. `esl-epfl/x-heep`
   Similar to 3.

8. `mohanson/accu`
   Documentation of a `bfp` instruction.

So, I see the following use cases:
* Old scalar crypto intrinsics (obsolete)
* Documentation of instructions (does not depend on a supported data type)
* CSR access

For CSR access, there are the following options to write common CSR-wrappers for RV32 and RV64:
* use `unsigned long` (like Linux does in `arch/riscv/include/asm/csr.h`).
* define `uint_xlen_t` as needed (we expose `__riscv_xlen` which makes this trivial)

@Timmmm, I don't know your use case. Could you please explain why these options don't fit for you?

There was no use of `int_xlen_t` in the search results, so we have to justify defining a type with no known use case (or drop defining this type).

Regarding the name `uint_xlen_t`, it was already mentioned above that defining a `uint_`-prefixed type collides with the C99 data type naming and is misleading, as it hides the fact that this is RISC-V-specific (it was proposed to use `__riscv_uint_xlen_t` instead).

Beyond that, it needs to be decided where to define this type and the supporting macros (e.g. libc headers, compiler, intrinsics header). I cannot see a clear solution here, and I could not find a precedent for arch-specific data types, so this could become a heated discussion with all stakeholders (upstream maintainers of all affected projects).

So far, nobody has cared enough to solve these problems just to tweak CSR wrappers. Beyond that, I could only find intrinsics as possible use case for `uint_xlen_t`. However, since we went ahead with "32"/"64" intrinsics with explicit use of `uint32_t`/`uint64_t`, it is unlikely that intrinsics will be a future use case for `uint_xlen_t`.

cmuellner:
> > We discussed a similar proposal yesterday: the introduction of pseudo instructions for xlen-wide loads/stores.
> > While there are use cases that benefit from abstracting RV32/RV64 differences, these use cases are rare.
> > Therefore, the costs of introducing a data type and all related support exceed the benefits.
> 
> I think we may have accidentally commingled two things here. I had made the point in the toolchain SIG that the community of folks hand writing assembly that needs to run on both rv32 and rv64 is likely quite small, and that adding complexity to all tool-chain implementations would be hard to justify under that assumption. (To be clear here - "hard" does not mean "impossible"; it just means significant justification is needed.)
> 
> My point (which is what I think you're referencing here) was _not_ in relation to whether we wanted/needed support for C/C++ language source which is compiled to both rv32 and rv64. That's a different, possibly much larger, user community. To be fair, the discussion on this issue to date doesn't seem to make that distinction itself very well - as the major use case mentioned is in fact interacting with inline assembly.
> 
> To be clear, I am not speaking in support of this proposal; I currently have no opinion on its merits. I'm only noting that the justification required for a C/C++ feature and an assembler feature are _different_ and we should be careful not to commingle them,
> 
> If this feature does have a motivating use which is not interacting with inline assembly, it would be good if the advocates would explicitly spell it out.

I haven't stated that you commented on this topic during the last SIG Toolchain meeting.
And I haven't used your comment on the other topic as motivation to close this ticket.
Sorry if I made that impression.

This ticket was discussed in 2020, 2023 and now, 2025. There is sympathy for having a `uint_xlen_t` data type, but nobody cared enough so far to address all the concerns this raises (they are stated in this ticket). Therefore, I've closed it as not planned.

Timmmm:
> * use unsigned long (like Linux does in arch/riscv/include/asm/csr.h).
> * define uint_xlen_t as needed (we expose __riscv_xlen which makes this trivial)
> 
> @Timmmm, I don't know your use case. Could you please explain why these options don't fit for you?

It's CSR access. I guess we can use `unsigned long`; I just don't like using those native C types since they can be crazy sizes (nothing technically stopping it being 64 bits on RV32).

I've just gone with [this](https://github.com/riscv/sail-riscv/blob/ace45d506d8da6d72da77ab1dce1d7cf07fd1871/test/first_party/src/common/runtime.h#L7-L54).

PR #16, opened 2021-01-11
Correct example definition of __riscv_f

simonpcook:


kito-cheng:
Oh, thanks for fix that :)

jim-wilson (approved):

PR #22, opened 2021-11-05
Add __riscv_vector back from deprecated

eopXD:
After adding the zve extension it is not obvious for user whether if there is
any vector instructions available. Adding __riscv_vector will let the users
able to check this.


kito-cheng:
LGTM, but I guess we need gather few more LGTM from LLVM community.

@jrtc27 @asb @luismarques @rdolbeau @nick-knight @rjiejie tag some people who might interested.

rjiejie:
LGTM.

asb:
I think adding this has the same justification as Craig's recent refactoring in the LLVM backend to [add hasStdExtV](https://reviews.llvm.org/D112496). It makes sense to me.

jim-wilson (approved):

eopXD:
Rebased to latest main.

PR #23, opened 2021-11-05
scalar crypto v1.0 builtins / intrinsics

mjosaarinen:
Derived from https://github.com/rvkrypto/rvkrypto-fips/blob/main/rvkintrin.md 
A serious attempt has been made to match with Scalar Cryptography v1.0rc5 specification.
Submitted as a "design document" so that GCC and LLVM may have equivalent features implementations.

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -58,19 +58,34 @@ For example:
 - F-extension v2.2 will define `__riscv_f` as `2002000`.
 - B-extension v0.92 will define `__riscv_b` as `92000`.
 
-| Name                    | Value        | When defined                  |
-| ----------------------- | ------------ | ----------------------------- |
-| __riscv_arch_test       | 1            | Defined if compiler support new architecture extension test macro. |
-| __riscv_i               | Arch Version | `I` extension is available.   |
-| __riscv_e               | Arch Version | `E` extension is available.   |
-| __riscv_m               | Arch Version | `M` extension is available.   |
-| __riscv_a               | Arch Version | `A` extension is available.   |
-| __riscv_f               | Arch Version | `F` extension is available.   |
-| __riscv_d               | Arch Version | `D` extension is available.   |
-| __riscv_c               | Arch Version | `C` extension is available.   |
-| __riscv_b               | Arch Version | `B` extension is available.   |
-| __riscv_v               | Arch Version | `V` extension is available.   |
-| __riscv_zfh             | Arch Version | `Zfh` extension is available. |
+
+| Name                    | Value        | When defined                                                 |
+| ----------------------- | ------------ | ------------------------------------------------------------ |
+| __riscv_arch_test       | 1            | Compiler supports architecture extension test macros.        |
+| __riscv_i               | Arch Version | `I` extension is available.                                  |
+| __riscv_e               | Arch Version | `E` extension is available.                                  |
+| __riscv_m               | Arch Version | `M` extension is available.                                  |
+| __riscv_a               | Arch Version | `A` extension is available.                                  |
+| __riscv_f               | Arch Version | `F` extension is available.                                  |
+| __riscv_d               | Arch Version | `D` extension is available.                                  |
+| __riscv_c               | Arch Version | `C` extension is available.                                  |
+| __riscv_b               | Arch Version | `B` extension is available.                                  |
+| __riscv_v               | Arch Version | `V` extension is available.                                  |
+| __riscv_zfh             | Arch Version | `Zfh` extension is available.                                |
+| __riscv_zk              | Arch Version | Extensions of `Zkn` are available and `Zkt` is asserted.     |
+| __riscv_zkn             | Arch Version | Zkn: `Zbkb` `Zbkc` `Zbkx` `Zkne` `Zknd` `Zknh` are all available. |
+| __riscv_zks             | Arch Version | Zks: `Zbkb` `Zbkc` `Zbkx` `Zksed` `Zksh` are all available.       |
+| __riscv_zbkb            | Arch Version | `Zbkb` extension is available.                               |
+| __riscv_zbkc            | Arch Version | `Zbkc` extension is available.                               |
+| __riscv_zbkx            | Arch Version | `Zbkx` extension is available.                               |
+| __riscv_zknd            | Arch Version | `Zknd` extension is available.                               |
+| __riscv_zkne            | Arch Version | `Zkne` extension is available.                               |
+| __riscv_zknh            | Arch Version | `Zknh` extension is available.                               |
+| __riscv_zksed           | Arch Version | `Zksed` extension is available.                              |
+| __riscv_zksh            | Arch Version | `Zksh` extension is available.                               |
+| __riscv_zkt             | Arch Version | Target asserts `Zkt` (data-independent latency extension).   |
+| __riscv_zkr             | Arch Version | `Zkr`  extension is available.                               |
+
 
 ### ABI Related Preprocessor Definitions
 
@@ -114,7 +129,96 @@ For example:
 * `__attribute__((interrupt("supervisor")))`
 * `__attribute__((interrupt("machine")))`
 
-## Intrinsic Functions
 
-Do we really have none of these?  I can't figure out
-`gcc/gcc/config/riscv/riscv-builtins.c`...
+## Scalar Cryptography Extension Intrinsics
+
+In order to access the RISC-V scalar crypto intrinsics, it is necessary to
+include the header file `riscv_crypto.h`.
+
+The functions are only only available if the compiler's `-march` string
+enables the required ISA extension. (Calling functions for not enabled
+ISA extensions will lead to compile-time and/or link-time errors.)
+
+The prefixes and data types are:
+
+* `_rv_*(...)`: intrinsics that operate on the `long` data type.
+* `_rv32_*(...)`: intrinsics that operate on the `int32_t` data type.
+* `_rv64_*(...)`: RV64-only intrinsics that operate on the `int64_t` data type.
+
+The extensions `Zkt` or `Zkr` do not currently introduce intrinsics.
+Data-independent latency is a property of the implementation, while
+`Zkr` (the entropy source interface) is implemented as a CSR only
+(`seed` at CSR address `0x015`).
+
+Many of the `Zbkb` instructions can be inferred compilers so they do not
+have direct intrinsics equivalents. Rotations are included mainly as RISC-V
+sign extension behavior, and clamping of rotation amount may generate extra
+instructions unless the programmer is very careful.
+
+The `_rv64_sha512*` intrinsics are marked with optional [RV32] support.
+This will require the compiler to decompose them into pairs of the
+corresponding `_rv32_sha512*` instructions.
+
+### Scalar Cryptography Extension short form intrinsics (alphabetically)
+
+| Prototype                                                     | Mnemonic      | Short Description                         | Supported in                  |
+| ------------------------------------------------------------- | ------------- | ----------------------------------------- | ----------------------------- |
+| `int32_t _rv32_aes32dsi(int32_t rs1, int32_t rs2, int bs);`   | `aes32dsi`    | AES final round decryption / RV32.        | Zknd, Zkn, Zk (RV32)          |
+| `int32_t _rv32_aes32dsmi(int32_t rs1, int32_t rs2, int bs);`  | `aes32dsmi`   | AES middle round decryption / RV32.       | Zknd, Zkn, Zk (RV32)          |
+| `int32_t _rv32_aes32esi(int32_t rs1, int32_t rs2, int bs);`   | `aes32esi`    | AES final round encryption / RV32.        | Zkne, Zkn, Zk (RV32)          |
+| `int32_t _rv32_aes32esmi(int32_t rs1, int32_t rs2, int bs);`  | `aes32esmi`   | AES middle round encryption / RV32.       | Zkne, Zkn, Zk (RV32)          |
+| `int64_t _rv64_aes64ds(int64_t rs1, int64_t rs2);`            | `aes64ds`     | AES final round decryption / RV64.        | Zknd, Zkn, Zk (RV64)          |
+| `int64_t _rv64_aes64dsm(int64_t rs1, int64_t rs2);`           | `aes64dsm`    | AES middle round decryption / RV64        | Zknd, Zkn, Zk (RV64)          |
+| `int64_t _rv64_aes64es(int64_t rs1, int64_t rs2);`            | `aes64es`     | AES final round encryption / RV64.        | Zkne, Zkn, Zk (RV64)          |
+| `int64_t _rv64_aes64esm(int64_t rs1, int64_t rs2);`           | `aes64esm`    | AES middle round encryption / RV64.       | Zkne, Zkn, Zk (RV64)          |
+| `int64_t _rv64_aes64im(int64_t rs1);`                         | `aes64im`     | AES Inverse MixColumns, key schedule.     | Zknd, Zkn, Zk (RV64)          |
+| `int64_t _rv64_aes64ks1i(int64_t rs1, int rnum);`             | `aes64ks1i`   | AES key schedule, round number.           | Zkne, Zknd, Zkn, Zk (RV64)    |
+| `int64_t _rv64_aes64ks2(int64_t rs1, int64_t rs2);`           | `aes64ks2`    | AES key schedule, word mixing.            | Zkne, Zknd, Zkn, Zk (RV64)    |
+| `int32_t _rv32_brev8(int32_t rs1);`                           | `brev8`       | Reverse order of bits within each byte.   | Zbkb (RV32)                   |
+| `int64_t _rv64_brev8(int64_t rs1);`                           | `brev8`       | Reverse order of bits within each byte.   | Zbkb (RV64)                   |
+| `int32_t _rv32_clmul(int32_t rs1, int32_t rs2);`              | `clmul`       | Carry-less multiply (low 32 bits).        | Zbc, Zbkc (RV32)              |
+| `int64_t _rv64_clmul(int64_t rs1, int64_t rs2);`              | `clmul`       | Carry-less multiply (low 64 bits).        | Zbc, Zbkc (RV64)              |
+| `int32_t _rv32_clmulh(int32_t rs1, int32_t rs2);`             | `clmulh`      | Carry-less multiply (high 32 bits).       | Zbc, Zbkc (RV32)              |
+| `int64_t _rv64_clmulh(int64_t rs1, int64_t rs2);`             | `clmulh`      | Carry-less multiply (high 64 bits).       | Zbc, Zbkc (RV64)              |
+| `int32_t _rv32_rol(int32_t rs1, int32_t rs2);`                | `rol[i][w]`   | Circular left rotate of 32 bits.          | Zbb, Zbkb (RV32,RV64)         |
+| `int64_t _rv64_rol(int64_t rs1, int64_t rs2);`                | `rol`/`rori`  | Circular left rotate of 64 bits.          | Zbb, Zbkb (RV64)              |
+| `int32_t _rv32_ror(int32_t rs1, int32_t rs2);`                | `ror[i][w]`   | Circular right rotate of 32 bits.         | Zbb, Zbkb (RV32,RV64)         |
+| `int64_t _rv64_ror(int64_t rs1, int64_t rs2);`                | `ror[i]`      | Circular right rotate of 64 bits.         | Zbb, Zbkb (RV64)              |
+| `long _rv_sha256sig0(long rs1);`                              | `sha256sig0`  | Sigma0 function for SHA2-256.             | Zknh, Zkn, Zk (RV32,RV64)     |
+| `long _rv_sha256sig1(long rs1);`                              | `sha256sig1`  | Sigma1 function for SHA2-256.             | Zknh, Zkn, Zk (RV32,RV64)     |
+| `long _rv_sha256sum0(long rs1);`                              | `sha256sum0`  | Sum0 function for SHA2-256.               | Zknh, Zkn, Zk (RV32,RV64)     |
+| `long _rv_sha256sum1(long rs1);`                              | `sha256sum1`  | Sum1 function for SHA2-256.               | Zknh, Zkn, Zk (RV32,RV64)     |
+| `int32_t _rv32_sha512sig0h(int32_t rs1, int32_t rs2);`        | `sha512sig0h` | Sigma0 high half for SHA2-512.            | Zknh, Zkn, Zk (RV32)          |
+| `int32_t _rv32_sha512sig0l(int32_t rs1, int32_t rs2);`        | `sha512sig0l` | Sigma0 low half for SHA2-512.             | Zknh, Zkn, Zk (RV32)          |
+| `int32_t _rv32_sha512sig1h(int32_t rs1, int32_t rs2);`        | `sha512sig1h` | Sigma1 high half for SHA2-512.            | Zknh, Zkn, Zk (RV32)          |
+| `int32_t _rv32_sha512sig1l(int32_t rs1, int32_t rs2);`        | `sha512sig1l` | Sigma1 low half for SHA2-512.             | Zknh, Zkn, Zk (RV32)          |
+| `int32_t _rv32_sha512sum0r(int32_t rs1, int32_t rs2);`        | `sha512sum0r` | Sum0 function for SHA2-512.               | Zknh, Zkn, Zk (RV32)          |
+| `int32_t _rv32_sha512sum1r(int32_t rs1, int32_t rs2);`        | `sha512sum1r` | Sum1 function for SHA2-512.               | Zknh, Zkn, Zk (RV32)          |
+| `int64_t _rv64_sha512sig0(int64_t rs1);`                      | `sha512sig0`  | Sigma0 function for SHA2-512.             | Zknh, Zkn, Zk (RV64 [RV32])   |
+| `int64_t _rv64_sha512sig1(int64_t rs1);`                      | `sha512sig1`  | Sigma1 function for SHA2-512.             | Zknh, Zkn, Zk (RV64 [RV32])   |
+| `int64_t _rv64_sha512sum0(int64_t rs1);`                      | `sha512sum0`  | Sum0 function for SHA2-512.               | Zknh, Zkn, Zk (RV64 [RV32])   |
+| `int64_t _rv64_sha512sum1(int64_t rs1);`                      | `sha512sum1`  | Sum1 function for SHA2-512.               | Zknh, Zkn, Zk (RV64 [RV32])   |
+| `long _rv_sm3p0(long rs1);`                                   | `sm3p0`       | P0 function for SM3 hash.                 | Zksh, Zks (RV32,RV64)         |
+| `long _rv_sm3p1(long rs1);`                                   | `sm3p1`       | P1 function for SM3 hash.                 | Zksh, Zks (RV32,RV64)         |
+| `long _rv_sm4ed(int32_t rs1, int32_t rs2, int bs);`           | `sm4ed`       | Accelerate SM4 cipher encrypt/decrypt.    | Zksed, Zks (RV32,RV64)        |
+| `long _rv_sm4ks(int32_t rs1, int32_t rs2, int bs);`           | `sm4ks`       | Accelerate SM4 cipher key schedule.       | Zksed, Zks (RV32,RV64)        |
+| `int32_t _rv32_unzip(int32_t rs1);`                           | `unzip`       | Odd/even bits into upper/lower halves.    | Zbkb (RV32)                   |
+| `int32_t _rv32_xperm4(int32_t rs1, int32_t rs2);`             | `xperm4`      | Nibble-wise lookup of indices.            | Zbkx (RV32)                   |
+| `int64_t _rv64_xperm4(int64_t rs1, int64_t rs2);`             | `xperm4`      | Nibble-wise lookup of indices.            | Zbkx (RV64)                   |
+| `int32_t _rv32_xperm8(int32_t rs1, int32_t rs2);`             | `xperm8`      | Byte-wise lookup of indices.              | Zbkx (RV32)                   |
+| `int64_t _rv64_xperm8(int64_t rs1, int64_t rs2);`             | `xperm8`      | Byte-wise lookup of indices.              | Zbkx (RV64)                   |
+| `int32_t _rv32_zip(int32_t rs1);`                             | `zip`         | Upper/lower halves into odd/even bits.    | Zbkb (RV32)                   |
+
+### Cryptography Intrinsics Implementation Guarantees
+
+The `riscv_crypto.h` can implement the intrinsics in many ways
+(early implementations used inline assembler). Builtin mapping is a
+compiler and system specific issue.
+
+Due to the data-independent latency ("constant time") assertions of
+the `Zkt` extension, the header file or the compiler can't use table
+lookups, conditional branching, etc., when implementing crypto intrinsics.
+However, this approach has been used for functional validation.
+In production (cryptographic implementations), the execution latency of
+all cryptography intrinsics must be independent of input values.
+


riscv-c-api.md
```diff
@@ -114,7 +114,221 @@ For example:
...
+
+The extensions `Zkt` or `Zkr` do not introduce builtins and do not have
+architecture extension test macros.
+
+However, **all** of the scalar cryptography builtins below must offer the
+same data-independent latency guarantees as the `Zkt` extension. The user
+must be able to trust that the compiler can't use table lookups, conditional
+branching, etc when transforming these crypto builtins into code.
+
+Shorter intrinsics can be be mapped to builtins via the `rvkintrin.h` system
```
kito-cheng, line 148: `rvkintrin.h` seems only mention here, but no detail info, I would we suggest define which API/intrinsic has exposed in this header instead of defining low-level built-in function.

riscv-c-api.md
```diff
@@ -114,7 +114,221 @@ For example:
...
+| __riscv_zbkc  | Arch Version | `Zbkc` extension is available.                               |
+| __riscv_zbkx  | Arch Version | `Zbkx` extension is available.                               |
+| __riscv_zknd  | Arch Version | `Zknd` extension is available.                               |
+| __riscv_zkne  | Arch Version | `Zkne` extension is available.                               |
+| __riscv_zknh  | Arch Version | `Zknh` extension is available.                               |
+| __riscv_zksed | Arch Version | `Zksed` extension is available.                              |
+| __riscv_zksh  | Arch Version | `Zbsh` extension is available.                               |
+
+The extensions `Zkt` or `Zkr` do not introduce builtins and do not have
+architecture extension test macros.
```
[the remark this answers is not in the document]
mjosaarinen: Zkr is a CSR so it's not a compiler matter. Added Zkt.

riscv-c-api.md
```diff
@@ -114,7 +114,221 @@ For example:
...
+| __riscv_zkn   | Arch Version | `Zbkb` `Zbkc` `Zbkx` `Zkne` `Zknd` `Zknh` are all available. |
+| __riscv_zks   | Arch Version | `Zbkb` `Zbkc` `Zbkx` `Zksed` `Zksh` are all available.       |
+| __riscv_zbkb  | Arch Version | `Zbkb` extension is available.                               |
+| __riscv_zbkc  | Arch Version | `Zbkc` extension is available.                               |
+| __riscv_zbkx  | Arch Version | `Zbkx` extension is available.                               |
+| __riscv_zknd  | Arch Version | `Zknd` extension is available.                               |
+| __riscv_zkne  | Arch Version | `Zkne` extension is available.                               |
+| __riscv_zknh  | Arch Version | `Zknh` extension is available.                               |
+| __riscv_zksed | Arch Version | `Zksed` extension is available.                              |
+| __riscv_zksh  | Arch Version | `Zbsh` extension is available.                               |
```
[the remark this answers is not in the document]
mjosaarinen: Merged.

riscv-c-api.md
mjosaarinen: Added the file from https://github.com/rvkrypto/rvkrypto-fips/  that defineds a mapping. That code has been significantly refactored now.

riscv-c-api.md
```diff
@@ -114,7 +114,221 @@ For example:
...
+
+The extensions `Zkt` or `Zkr` do not introduce builtins and do not have
+architecture extension test macros.
+
+However, **all** of the scalar cryptography builtins below must offer the
+same data-independent latency guarantees as the `Zkt` extension. The user
+must be able to trust that the compiler can't use table lookups, conditional
+branching, etc when transforming these crypto builtins into code.
```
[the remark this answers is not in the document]
mjosaarinen: Well, cryptographic code needs to have execution latency that is independent of secret values being processed; a property commonly (somewhat misleadingly) called "constant-time code". This is an assumption that e.g. the Kernel and OpenSSL make in relation to code generation, and it should be pretty automatically satisfied. Note that there are some instructions in addition to load/store and branches which are _not_ assumed to be constant time; e.g. division and remaindering. 

Please see Chapter 5 of the crypto spec for more info on the constant-time assertion (`Zkt` extension). I just basically reworded the thing:  _"Due to the data-independent latency ("constant time") assertions of the Zkt extension, the compiler can't use table lookups, conditional branching, etc when transforming these crypto intrinsics or builtins into code. Execution latency for them must be independent of input values."_

riscv-c-api.md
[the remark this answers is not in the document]
kito-cheng: `zicsr` also has define the architecture extension test macros, and those marco also could used for writing assembly file(*.S) file

riscv-c-api.md
```diff
@@ -114,7 +127,209 @@ For example:
...
+int64_t __builtin_riscv_xperm4_64(int64_t rs1, int64_t rs2);    //  XPERM4
+int64_t __builtin_riscv_xperm8_64(int64_t rs1, int64_t rs2);    //  XPERM8
+```
+
+### Zknd (Zk, Zkn):     NIST Suite: AES Decryption
+
+```C
+//  Zk, Zkn, Zknd on RV32
+int32_t __builtin_riscv_aes32dsi(int32_t rs1, int32_t rs2, int bs);     //  AES32DSI
+int32_t __builtin_riscv_aes32dsmi(int32_t rs1, int32_t rs2, int bs);    //  AES32DSMI
```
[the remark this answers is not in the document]
mjosaarinen: The legal values are { 0, 1, 2, 3 } and that input must be a constant immediate (never a register -- i.e. `int32_t`). `int` is the default "native" type for a numerical constant in the C language, so it seems appropriate to me?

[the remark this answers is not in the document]
mjosaarinen: The inline assembler (non-builtin) instantiation at https://github.com/rvkrypto/rvkrypto-fips/blob/2f03b0d155415be0b1396682b1d365c1a522b34e/rvk_asm_intrin.h#L122
```C
static inline int32_t _rvk_asm_aes32dsi(int32_t rs1, int32_t rs2, int bs)
        { int32_t rd; __asm__("aes32dsi  %0, %1, %2, %3" : "=r"(rd) : "r"(rs1), "r"(rs2), "i"(bs)); return rd; }
static inline int32_t _rvk_asm_aes32dsmi(int32_t rs1, int32_t rs2, int bs)
        { int32_t rd; __asm__("aes32dsmi  %0, %1, %2, %3" : "=r"(rd) : "r"(rs1), "r"(rs2), "i"(bs)); return rd; }
```

alextsao1999:
Hi there,
I'm implementing the rvk intrinsics, but I'm confused with the differences between int and int32_t.
```c
//  Zk, Zkn, Zkne on RV32
int32_t __builtin_riscv_aes32esi(int32_t rs1, int32_t rs2, int bs);     //  AES32ESI
int32_t __builtin_riscv_aes32esmi(int32_t rs1, int32_t rs2, int bs);    //  AES32ESMI
```

riscv-c-api.md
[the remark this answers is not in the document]
ChunyuLiao: thanks!

cmuellner:
We have discussed this PR in today's T&R meeting and we would like to move this forward.
Kito already provided some comments from the meeting.

Further ideas:
* Don't use 32-bit and 64-bit versions of intrinsics. The compilers support function overloading (e.g. `type __my_intrinsic (type *ptr)`)
* Propose generic (non-RV specific) intrinsics for crypto primitives like AES enc/dec or SHA calculation as generic crypto intrinsics

Some inspiration can be found here: [GCC's atomic builtins API](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html)

mjosaarinen:
Kito:
> -   Define the header file name for intrinsic and the intrinsic function list for name and semantic.

Gmm.. So you're asking for a header that just contains prototypes?

The current `rvkintrin.h` header file maps to builtins, inline assembler, or emulation, depending on the environment  https://github.com/rvkrypto/rvkrypto-fips/blob/main/rvkintrin.h

> -   Some intrinsic is compiler's generic builtin like __builtin_bswap32, those intrinsic don't need to list here.

Yeah, there is no need to have it in this header file. Note that you don't need to enable Bitmanip extension to use REV8 to implement __builtin_bswap32, scalar crypto is sufficient.

> -   Few instructions can be synthesis by compiler like andn, orn and xnor, those intrinsic don't need to list too.

Yep, those don't need to be explicit.

Christoph:
> Further ideas:
> 
>     * Don't use 32-bit and 64-bit versions of intrinsics. The compilers support function overloading (e.g. `type __my_intrinsic (type *ptr)`)

Hmm.. Can you give concrete examples of where this should be done?

Note that rotations, for example, have different behavior on 32- bit and 64- bit targets, and furthermore both are available on RV64 (ROR and RORW). So the bit size needs to be explicitly stated.

For most cryptographic intrinsics, the behavior of 32-bit and 64-bit versions is completely different -- even though they have been designed to achieve the same goal. So sharing this would create confusion (and dangerous, broken cryptography if it is possible to compile code designed for 32-bit targets on a 64-bit system, or vice versa).

The way primitives such as AES or SHA2-512, SHA3 are implemented using the intrinsics differs completely between RV32 and RV64. The intrinsics for for SHA2-256 and SM3, SM4 are fundamentally "32-bit" on both targets and hence the same names are already used.

> 
>     * Propose generic (non-RV specific) intrinsics for crypto primitives like AES enc/dec or SHA calculation as generic crypto intrinsics

The scalar cryptographic extension does not have instructions for full AES encryption, decryption, or SHA calculation. Such macro primitives should be located in libraries. 

No other architecture has a scalar cryptography extension, so the generality would be questionable in this particular case.

Cheers,
- markku

cmuellner:
Hi Markku,

> Christoph:
> 
> > Further ideas:
> > ```
> > * Don't use 32-bit and 64-bit versions of intrinsics. The compilers support function overloading (e.g. `type __my_intrinsic (type *ptr)`)
> > ```
> 
> Hmm.. Can you give concrete examples of where this should be done?
> 
> Note that rotations, for example, have different behavior on 32- bit and 64- bit targets, and furthermore both are available on RV64 (ROR and RORW). So the bit size needs to be explicitly stated.

I think this could be applied to `ror`, `rol`, `pack`, `rev8`, and `clmul` (unless this would cause ambiguity).
E.g. we could have these two intrinsics for rotations (inspired by a [C++ standard proposal](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0553r4.html)):
```
T __builtin_rotl(T val, int n);
T __builtin_rotr(T val, int n);
```

The compiler can then emit a `rorw` for a code like `s64 val64 = (s64)__builtin_rotr(val_in, n);` by consuming 32-bit zero-extension of `val_in`, the 32-rotation and the following sign-extension to 64-bit.

> For most cryptographic intrinsics, the behavior of 32-bit and 64-bit versions is completely different -- even though they have been designed to achieve the same goal. So sharing this would create confusion (and dangerous, broken cryptography if it is possible to compile code designed for 32-bit targets on a 64-bit system, or vice versa).

I see your point. However, I would expect implementations to use correct 32-bit or 64-bit data types in the code (C99 is soon 23 years there). So the code can be compiled for 32-bit and 64-bit targets, but the resulting instructions might differ. But if the code does not compile for 32-bit targets, because it has been written for 64-bit targets (using `__builtin_riscv_ror_64`), then somebody will have to touch this code and potentially break it.

> The way primitives such as AES, SHA2-512, or SHA3 are implemented using the intrinsics differs completely between RV32 and RV64. The intrinsics for for SHA2-256 and SM3, SM4 are fundamentally "32-bit" on both targets and hence the same names are already used.

Yes, I agree that the above idea does not apply here.

> > ```
> > * Propose generic (non-RV specific) intrinsics for crypto primitives like AES enc/dec or SHA calculation as generic crypto intrinsics
> > ```
> 
> The scalar cryptographic extension does not have instructions for full AES encryption, decryption, or SHA calculation. Such macro primitives should be located in libraries in my opinion.
> 
> No other architecture has a scalar cryptography extension, so the generality would be questionable in this particular case.

I could imagine something like `aes_{N}_{op}_key_schedule` and `aes_{N}_ecb_{op}rypt` for AES.
This would help to propagate the Zk* support in crypto libraries.
However, I agree that this proposal implies some challenges (which algorithms, which API abstraction, etc).

kito-cheng:
> Hmm.. So you're asking for a header that just contains prototypes?
>
> The current rvkintrin.h header file maps to builtins, inline assembler, or emulation, depending on the environment https://github.com/rvkrypto/rvkrypto-fips/blob/main/rvkintrin.h

The spec only define prototypes (interface and behavior), but let the compiler have freedom to decide how to implement that.

And inline asm and emulation version should not included in intrinsic header file in my opinion, that should just used for evaluation only.

mjosaarinen:
Yep.. asm and emulation are just for evaluation, and a plain header file is easy enough to extract. However, these functions don't have linkage and I"m not so sure how the prototypes are used or where such a header file should go.

topperc:
clang has `__builtin_rotateright32`, `__builtin_rotateright64`, `__builtinrotateleft32` and `__builtin_rotateleft64`. These expand to the shifts and or when they aren't supported natively. So no target specific intrinsics are needed for clang.

I think the `__builtin` interface should be considered a compiler implementation detail and not specified. Only the _rv, _rv32, _rv64 names should be specified. A compiler could choose to recognize the _rv, _rv32, _rv64 function names as special symbols whenever it has seen the an include directive for the header if it so chose.

mjosaarinen:
After discussing this with Christoph today, I took the action to remove the Zkbk intrinsics for ANDN, ORN, XNOR, PACK[H], PACKW, REV8 (bswap32) as they can be inferred by the compiler in most cases.

I also added an alphabetic list with short descriptions of the intrinsics. The bultins are now merely suggested and basically an implementation matter. From a programming viewpoint, I'll want to include the header anyway and the platform can map that to the appropriate functions how it sees appropriate.

I didn't remove `_rv32/64_ror()` and friends yet since the clang  `__builtin_rotateright32`, `__builtin_rotateright64`, `__builtinrotateleft32` and `__builtin_rotateleft64` are not in GCC documentation. It makes sense to remove platform-specific builtins and intrinsics only if there is a universal replacement.

topperc:
> After discussing this with Christoph today, I took the action to remove the Zkbk intrinsics for ANDN, ORN, XNOR, PACK[H], PACKW, REV8 (bswap32) as they can be inferred by the compiler in most cases.

I think it may still make sense to have an `__rv` named bswap function. The header file implementation for clang and gcc could be to implement it with __builtin_bswap32.

> 
> I also added an alphabetic list with short descriptions of the intrinsics. The bultins are now merely suggested and basically an implementation matter. From a programming viewpoint, I'll want to include the header anyway and the platform can map that to the appropriate functions how it sees appropriate.
> 
> I didn't remove `_rv32/64_ror()` and friends yet since the clang `__builtin_rotateright32`, `__builtin_rotateright64`, `__builtinrotateleft32` and `__builtin_rotateleft64` are not in GCC documentation. It makes sense to remove platform-specific builtins and intrinsics only if there is a universal replacement.

I think it makes sense to keep the `_rv32/64_ror()` and friends interface. Clang can implement them in terms of it's `__builtin_rotateright32`, `__builtin_rotateright64`, `__builtinrotateleft32` and `__builtin_rotateleft64`. Assuming we consider builtin names to be more of an implementation detail and don't require clang to implement RISCV specific builtins.

As RISCV gets more popular in the future, a C compiler other than gcc or llvm with RISCV support may emerge, it may not want to implement the __builtin* names. So having an interface for bswap and rotate defined by RISCV that doesn't depend on __builtin* makes sense to me.

cmuellner:
> As RISCV gets more popular in the future, a C compiler other than gcc or llvm with RISCV support may emerge, it may not want to implement the __builtin* names. So having an interface for bswap and rotate defined by RISCV that doesn't depend on __builtin* makes sense to me.

Understood.
Would you agree, that there is no need for `ANDN`, `ORN`, `XNOR` intrinsics, because they can be easily expressed in C (or an IR)?

What I would like to avoid is, that we define APIs, that don't have a valid use-case in modern code and therefore are used more or less only by accident. Therefore I prefer to have platform-independent APIs if the functionality is not very arch-specific. E.g. `T __builtin_rotr(T val, int n);` is better than `int32_t __builtin_riscv_ror_32(int32_t rs1, int32_t rs2);`. Do you see a good use-case where `__builtin_riscv_ror_32()` would be preferred?

One option would be to not define any intrinsics for rotations/bswap/etc. at all because we expect that this functionality is better offered by an arch-independent API. Also note, that we don't have any intrinsics for Zb* instructions so far.

mjosaarinen:
Hi,

I think it's better to call the things defined by the header as
"intrinsics" and keep the builtins as a compiler implementation matter
(that is merely suggested). This is what the PR updated yesterday does.

It would be better if you moved this discussion there and comment on the
actual PR so that we have a record.

Cheers,
Markku

On Fri, Feb 4, 2022, 00:53 Christoph Müllner ***@***.***>
wrote:

> As RISCV gets more popular in the future, a C compiler other than gcc or
> llvm with RISCV support may emerge, it may not want to implement the
> __builtin* names. So having an interface for bswap and rotate defined by
> RISCV that doesn't depend on __builtin* m sense to me.
>
> Understood.
> Would you agree, that there is no need for ANDN, ORN, XNOR intrinsics,
> because they can be easily expressed in C (or an IR)?
>
What I would like to avoid is, that we define APIs, that don't have a valid

topperc:
> > As RISCV gets more popular in the future, a C compiler other than gcc or llvm with RISCV support may emerge, it may not want to implement the __builtin* names. So having an interface for bswap and rotate defined by RISCV that doesn't depend on __builtin* makes sense to me.
> 
> Understood. Would you agree, that there is no need for `ANDN`, `ORN`, `XNOR` intrinsics, because they can be easily expressed in C (or an IR)?

I agree. Those are easy to match. There may be some improvements around comparisons that can be made with ORN that LLVM doesn't know yet. LLVM does know some tricks for ANDN, see https://reviews.llvm.org/D115922 for example.

> 
> What I would like to avoid is, that we define APIs, that don't have a valid use-case in modern code and therefore are used more or less only by accident. Therefore I prefer to have platform-independent APIs if the functionality is not very arch-specific. E.g. `T __builtin_rotr(T val, int n);` is better than `int32_t __builtin_riscv_ror_32(int32_t rs1, int32_t rs2);`. Do you see a good use-case where `__builtin_riscv_ror_32()` would be preferred?
> 
> One option would be to not define any intrinsics for rotations/bswap/etc. at all because we expect that this functionality is better offered by an arch-independent API. Also note, that we don't have any intrinsics for Zb* instructions so far.

gcc and LLVM should be able to recognize rotates from `(x << n) | (x >> ((32 - n) & 31))`. I think LLVM can in some cases mess up that idiom before it gets a chance to realize it's a rotate. That's one of the reasons __builtin_rotate* was added. It looks like we're having trouble recognizing 32-bit rotates from that idiom with Zbb on RV64. Probably some side effect of i32 being considered an illegal type.

For the ratified Zb* extensions, clang implements builtins for clmul* and orc.b. There is no header file defined.

kito-cheng (approved):
Otherwise LGTM :)

mjosaarinen:
Thanks kito..
* Keeping the name `rvkintrin.h`.
* Removed the builtin suggestions as those are only relevant to compiler writers in 2022, not the end-users of the intrinsics. Moved the implementation suggestions at the end to make the representation more compact.
* I didn't add the byte reversal intrinsic back in -- we hope compilers can figure it out. The 32-bit byte reversal on RV64 is two instructions anyway (`rev8` + `srai`) and the functionality is available in many other ways, including `__htonl()` and `__builtin_bswap32()`. For 64-bit `rev8` we have `__builtin_bswap64()`. 
* Kept the rotations with the explanation that inferring the specific RISC-V sign extension behavior (for `rorw`, `roriw` `rolw`) and clamping of rotation amount may generate some extra code unless the programmer is careful with casts and masking of shift amounts.

riscv-c-api.md
```diff
@@ -114,7 +127,88 @@ For example:
...
+| `int64_t _rv64_sha512sum1(int64_t rs1);`                      | `sha512sum1`  | Sum1 function for SHA2-512.               | Zknhv, Zknv, Zk (RV64)        |
+| `long _rv_sm3p0(long rs1);`                                   | `sm3p0`       | P0 function for SM3 hash.                 | Zksh, Zks (RV32,RV64)         |
+| `long _rv_sm3p1(long rs1);`                                   | `sm3p1`       | P1 function for SM3 hash.                 | Zksh, Zks (RV32,RV64)         |
+| `long _rv_sm4ed(int32_t rs1, int32_t rs2, int bs);`           | `sm4ed`       | Accelerate SM4 cipher encrypt/decrypt.    | Zksed, Zks (RV32,RV64)        |
+| `long _rv_sm4ks(int32_t rs1, int32_t rs2, int bs);`           | `sm4ed`       | Accelerate SM4 cipher key schedule.       | Zksed, Zks (RV32,RV64)        |
+| `int32_t _rv32_unzip(int32_t rs1);`                           | `unzip`       | Odd/even bits into upper/lower halves.    | Zbkb (RV32)                   |
+| `int32_t _rv32_xperm4(int32_t rs1, int32_t rs2);`             | `xperm4`      | Byte-wise lookup of indicies.             | Zbkx (RV32)                   |
```
topperc, line 192: This should be "Nibble" not "Byte" and xperm8 should be "Byte" instead of "Nibble"

riscv-c-api.md
```diff
@@ -114,7 +127,88 @@ For example:
...
+| `int64_t _rv64_clmulh(int64_t rs1, int64_t rs2);`             | `clmulh`      | Carry-less multiply (high 32 bits).       | Zbc, Zbkc (RV32)              |
+| `int32_t _rv32_clmulh(int32_t rs1, int32_t rs2);`             | `clmulh`      | Carry-less multiply (high 64 bits).       | Zbc, Zbkc (RV64)              |
+| `int32_t _rv32_rol(int32_t rs1, int32_t rs2);`                | `rol[i][w]`   | Circular left rotate of 32 bits.          | Zbb, Zbkb (RV32,RV64)         |
+| `int64_t _rv64_rol(int64_t rs1, int64_t rs2);`                | `rol`/`rori`  | Circular left rotate of 64 bits.          | Zbb, Zbkb (RV64)              |
+| `int32_t _rv32_ror(int32_t rs1, int32_t rs2);`                | `ror[i][w]`   | Circular right rotate of 32 bits.         | Zbb, Zbkb (RV32,RV64)         |
+| `int64_t _rv64_ror(int64_t rs1, int64_t rs2);`                | `ror[i]`      | Circular right rotate of 64 bits.         | Zbb, Zbkb (RV64)              |
+| `long _rv_sha256sig0(long rs1);`                              | `sha256sig0`  | Sigma0 function for SHA2-256.             | Zknhv, Zknv, Zk (RV32,RV64)   |
```
topperc, line 173: What is Zknhv? I only see Zknh in the spec.

riscv-c-api.md
mjosaarinen: Thanks.

cmuellner (changes requested):

rvkintrin.h
```diff
@@ -0,0 +1,278 @@
+//  rvkintrin.h
```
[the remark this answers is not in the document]
mjosaarinen: Yes, the inline assembler / emulation version can be found at: https://github.com/rvkrypto/rvkrypto-fips/blob/main/riscv_crypto.h

riscv-c-api.md
```diff
@@ -114,7 +127,88 @@ For example:
...
+| `int32_t _rv32_unzip(int32_t rs1);`                           | `unzip`       | Odd/even bits into upper/lower halves.    | Zbkb (RV32)                   |
+| `int32_t _rv32_xperm4(int32_t rs1, int32_t rs2);`             | `xperm4`      | Nibble-wise lookup of indicies.           | Zbkx (RV32)                   |
+| `int64_t _rv64_xperm4(int64_t rs1, int64_t rs2);`             | `xperm4`      | Nibble-wise lookup of indicies.           | Zbkx (RV64)                   |
+| `int32_t _rv32_xperm8(int32_t rs1, int32_t rs2);`             | `xperm8`      | Byte-wise lookup of indicies.             | Zbkx (RV32)                   |
+| `int64_t _rv64_xperm8(int64_t rs1, int64_t rs2);`             | `xperm8`      | Byte-wise lookup of indicies.             | Zbkx (RV64)                   |
+| `int32_t _rv32_zip(int32_t rs1);`                             | `zip`         | Upper/lower halves into odd/even bits.    | Zbkb (RV32)                   |
+
+### Cryptography Intrinsics Implementation Notes
+
+The `rvkintrin.h` can implement the intrinsics in many ways (early
```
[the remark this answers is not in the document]
mjosaarinen: Changed to Implementation Guarantees. It is important that the programmer knows that the compiler doesn't break constant time features etc.

riscv-c-api.md
```diff
@@ -114,7 +127,88 @@ For example:
...
+implementations used inline assembler). The builtin mapping is a
+compiler and system specific issue.
+
+* `_rv_xxx()` intrinsics may map to `__builtin_riscv_xxx()`.
+* `_rv32_xxx()` intrinsics may map to `__builtin_riscv_xxx_32()`.
+* `_rv64_xxx()` intrinsics may map to `__builtin_riscv_xxx_64()`.
+
+Due to the data-independent latency ("constant time") assertions of
```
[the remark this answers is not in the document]
mjosaarinen: Yep, it should stay that way; hence "guarantees." If the fact that all crypto intrinsics map to assembler instructions is not documented, what is there to stop someone from implementing them with sequences of instructions?

riscv-c-api.md
```diff
@@ -114,7 +127,88 @@ For example:
 * `__attribute__((interrupt("supervisor")))`
 * `__attribute__((interrupt("machine")))`
 
-## Intrinsic Functions
 
-Do we really have none of these?  I can't figure out
-`gcc/gcc/config/riscv/riscv-builtins.c`...
+## Scalar Cryptography Extension Intrinsics
+
+The short intrinsics are instantiated by including the `rvkintrin.h` system
```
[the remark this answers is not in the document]
mjosaarinen: Corrected a bit, added.

riscv-c-api.md
```diff
@@ -114,7 +127,88 @@ For example:
...
+header file.
+
+For post-ratification Scalar Cryptography v1.0, the Arch Version is `1000000`
```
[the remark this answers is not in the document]
mjosaarinen: Yeah, it seems redundant, given that clear rules for version numbers are given in the same document. Removed.

riscv-c-api.md
```diff
@@ -114,7 +127,88 @@ For example:
...
+The extensions `Zkt` or `Zkr` do not introduce intrinsics. The data-independent
+latency is a property of the implementation, while `Zkr` (the entropy source
+interface) is implemented as a CSR only (`seed` at CSR address `0x015`).
+
+Many of the `Zbkb` instructions can be inferred compilers so they do not
+have direct intrinsics equivalents. Rotations are included mainly as RISC-V
+sign extension behavior, and clamping of rotation amount may generate extra
+instructions unless the programmer is very careful.
+
+### Scalar Cryptography Extension short form intrinsics (alphabetically)
```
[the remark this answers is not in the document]
mjosaarinen: We don't have separate banner for "all intrinsics", so this is under the main Scalar Cryptography Extension Intrinsics header.

riscv-c-api.md
```diff
@@ -114,7 +127,88 @@ For example:
...
+
+| Prototype                                                     | Mnemonic      | Short Description                         | Supported in                  |
+| ------------------------------------------------------------- | ------------- | ----------------------------------------- | ----------------------------- |
+| `int32_t _rv32_aes32dsi(int32_t rs1, int32_t rs2, int bs);`   | `aes32dsi`    | AES final round decryption / RV32.        | Zknd, Zkn, Zk (RV32)          |
```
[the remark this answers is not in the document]
mjosaarinen: Added the basic rules back in:

"The prefixes and data types are:

* `_rv_*(...)`: intrinsics that operate on the `long` data type.
* `_rv32_*(...)`: intrinsics that operate on the `int32_t` data type.
* `_rv64_*(...)`: RV64-only intrinsics that operate on the `int64_t` data type.
 "

riscv-c-api.md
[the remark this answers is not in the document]
cmuellner: I see your point, but isn't it enough to require that the necessary zk* extension is enabled in the march string.
This guarantees that the compiler will emit the corresponding instruction (we can even include tests for that).
If the march string does not include the required extension for an intrinsic, then there will be a compiler or linker error.

riscv-c-api.md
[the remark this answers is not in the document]
cmuellner: See also: https://github.com/riscv-non-isa/riscv-c-api-doc/pull/25

mjosaarinen:
Yep, updated https://github.com/mjosaarinen/riscv-c-api-doc/blob/master/riscv-c-api.md

I also updated the prototype implementation with https://github.com/rvkrypto/rvkrypto-fips/commit/cf2b45526dbcada981577f761e6a296e66769a17   An actual `riscv_crypto.h` file is only in that repo now.

riscv-c-api.md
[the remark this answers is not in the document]
mjosaarinen: Perhaps I missed something, but there's nothing useful at https://github.com/riscv-non-isa/riscv-c-api-doc/pull/25 . The convention here is the one originally Claire defined for Bitmanip's "rvintrin.h", and which has been floating around for several years.

I personally don't want to define a new convention. The way I see it, If you want to define a new convention, then define a new convention, document it well, and explain why you did that. Don't expect other people to halt everything to try to guess what is it that you want.

riscv-c-api.md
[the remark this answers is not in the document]
mjosaarinen: That section is already gone.

riscv-c-api.md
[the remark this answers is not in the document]
cmuellner: Well, I guess you understand the common naming conventions have something useful.

It might have offended you that I did not reuse the ones that you are using. That was not my intention (apologies, if that is the case).

My proposal was based on the naming convention from the [vector intrinsics](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/rvv-intrinsic-rfc.md#naming-rules), because they are already implemented in upstream LLVM.

mjosaarinen:
@cmuellner wants to present new intrinsics to the CETG for approval.

riscv-c-api.md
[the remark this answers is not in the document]
mjosaarinen: Well, these are the common conventions we have used with the Scalar instructions. I did not invent them; I believe Claire did.

riscv-c-api.md
```diff
@@ -58,19 +58,33 @@ For example:
...
+| __riscv_zks             | Arch Version | `Zbkb` `Zbkc` `Zbkx` `Zksed` `Zksh` are all available.       |
+| __riscv_zbkb            | Arch Version | `Zbkb` extension is available.                               |
+| __riscv_zbkc            | Arch Version | `Zbkc` extension is available.                               |
+| __riscv_zbkx            | Arch Version | `Zbkx` extension is available.                               |
+| __riscv_zknd            | Arch Version | `Zknd` extension is available.                               |
+| __riscv_zkne            | Arch Version | `Zkne` extension is available.                               |
+| __riscv_zknh            | Arch Version | `Zknh` extension is available.                               |
+| __riscv_zksed           | Arch Version | `Zksed` extension is available.                              |
+| __riscv_zksh            | Arch Version | `Zksh` extension is available.                               |
+| __riscv_zkt             | Arch Version | Target asserts `Zkt` (data-independent latency extension).   |
```
topperc, line 86: Is Zkr missing?

riscv-c-api.md
```diff
@@ -58,19 +58,33 @@ For example:
...
+| __riscv_e               | Arch Version | `E` extension is available.                                  |
+| __riscv_m               | Arch Version | `M` extension is available.                                  |
+| __riscv_a               | Arch Version | `A` extension is available.                                  |
+| __riscv_f               | Arch Version | `F` extension is available.                                  |
+| __riscv_d               | Arch Version | `D` extension is available.                                  |
+| __riscv_c               | Arch Version | `C` extension is available.                                  |
+| __riscv_b               | Arch Version | `B` extension is available.                                  |
+| __riscv_v               | Arch Version | `V` extension is available.                                  |
+| __riscv_zfh             | Arch Version | `Zfh` extension is available.                                |
+| __riscv_zk              | Arch Version | `Zbkb` `Zbkc` `Zbkx` `Zkne` `Zknd` `Zknh` are all available. |
```
topperc, line 75: Should Zk include Zkr and Zkt?

riscv-c-api.md
mjosaarinen: Yes. The compiler wouldn't necessarily know if it is there or not, especially at compile time. Zkr does not introduce new instructions, but a CSR (called "seed", and having address 0x015), so it should be accessed with whatever builtin or intrinsic is used to issue the `csrrw rd, 0x015, x0` instruction. See test code at https://github.com/rvkrypto/rvkrypto-fips/blob/main/test/test_zkr.c

In a Linux system, under a typical configuration, the CSR is only available to the hypervisor or the Kernel, and hence it is of limited general programming interest.

riscv-c-api.md
mjosaarinen: Well, perhaps it should, but Zk automatically implies Zkt, as do all other cryptographic extensions -- like Zkbk alone. Zkr does not have any impact on the compiler. All cryptographic extensions are Zkt.

topperc: Why is Zkt only explicitly listed as implied by Zk in the spec?

mjosaarinen: You're right. Constant time execution is not implied by other extensions to anything but those specific instructions themselves; no Zkt. Zkt in Zk implies it for all the instructions in the list. https://github.com/mjosaarinen/riscv-c-api-doc/commit/27183d099a72560ea865b5443ae1333919324df4

mjosaarinen:
@topperc was still reviewing this, so reopening.

riscv-c-api.md
topperc: It looks like the LLVM implementation accepts Zkr in -march. And Zk in -march implies Zkr. This causes the __riscv_zkr to be emitted from the preprocecessor and it appears in elf attributes.

What is the desired behavior?

mjosaarinen: Yes, thinking about it, it's better to have it -- perhaps future versions of the Zkr extension may actually define ISA extensions for entropy / random numbers, not just a CSR. So I added `__riscv_zkr` in https://github.com/mjosaarinen/riscv-c-api-doc/commit/8c6c806549faa79632991d5fc78c4abb24e040f8

riscv-c-api.md
```diff
@@ -114,7 +129,92 @@ For example:
...
+| `int64_t _rv64_sha512sig1(int64_t rs1);`                      | `sha512sig1`  | Sigma1 function for SHA2-512.             | Zknh, Zkn, Zk (RV64)          |
+| `int64_t _rv64_sha512sum0(int64_t rs1);`                      | `sha512sum0`  | Sum0 function for SHA2-512.               | Zknh, Zkn, Zk (RV64)          |
+| `int64_t _rv64_sha512sum1(int64_t rs1);`                      | `sha512sum1`  | Sum1 function for SHA2-512.               | Zknh, Zkn, Zk (RV64)          |
+| `long _rv_sm3p0(long rs1);`                                   | `sm3p0`       | P0 function for SM3 hash.                 | Zksh, Zks (RV32,RV64)         |
+| `long _rv_sm3p1(long rs1);`                                   | `sm3p1`       | P1 function for SM3 hash.                 | Zksh, Zks (RV32,RV64)         |
+| `long _rv_sm4ed(int32_t rs1, int32_t rs2, int bs);`           | `sm4ed`       | Accelerate SM4 cipher encrypt/decrypt.    | Zksed, Zks (RV32,RV64)        |
+| `long _rv_sm4ks(int32_t rs1, int32_t rs2, int bs);`           | `sm4ed`       | Accelerate SM4 cipher key schedule.       | Zksed, Zks (RV32,RV64)        |
```
kito-cheng, line 200: Should be `sm4ks`?

riscv-c-api.md
```diff
@@ -114,7 +129,92 @@ For example:
...
+| `int32_t _rv32_unzip(int32_t rs1);`                           | `unzip`       | Odd/even bits into upper/lower halves.    | Zbkb (RV32)                   |
+| `int32_t _rv32_xperm4(int32_t rs1, int32_t rs2);`             | `xperm4`      | Nibble-wise lookup of indicies.           | Zbkx (RV32)                   |
+| `int64_t _rv64_xperm4(int64_t rs1, int64_t rs2);`             | `xperm4`      | Nibble-wise lookup of indicies.           | Zbkx (RV64)                   |
+| `int32_t _rv32_xperm8(int32_t rs1, int32_t rs2);`             | `xperm8`      | Byte-wise lookup of indicies.             | Zbkx (RV32)                   |
+| `int64_t _rv64_xperm8(int64_t rs1, int64_t rs2);`             | `xperm8`      | Byte-wise lookup of indicies.             | Zbkx (RV64)                   |
```
kito-cheng, line 205: Maybe `long _rv_xperm[4|8](long rs1, long rs2)` instead of `_rv32_xperm[4|8]` and `_rv64_xperm[4|8]`?

riscv-c-api.md
```diff
@@ -114,7 +129,92 @@ For example:
...
+| `int32_t _rv32_sha512sig1l(int32_t rs1, int32_t rs2);`        | `sha512sig1l` | Sigma1 low half for SHA2-512.             | Zknh, Zkn, Zk (RV32)          |
+| `int32_t _rv32_sha512sum0r(int32_t rs1, int32_t rs2);`        | `sha512sum0r` | Sum0 function for SHA2-512.               | Zknh, Zkn, Zk (RV32)          |
+| `int32_t _rv32_sha512sum1r(int32_t rs1, int32_t rs2);`        | `sha512sum1r` | Sum1 function for SHA2-512.               | Zknh, Zkn, Zk (RV32)          |
+| `int64_t _rv64_sha512sig0(int64_t rs1);`                      | `sha512sig0`  | Sigma0 function for SHA2-512.             | Zknh, Zkn, Zk (RV64)          |
+| `int64_t _rv64_sha512sig1(int64_t rs1);`                      | `sha512sig1`  | Sigma1 function for SHA2-512.             | Zknh, Zkn, Zk (RV64)          |
+| `int64_t _rv64_sha512sum0(int64_t rs1);`                      | `sha512sum0`  | Sum0 function for SHA2-512.               | Zknh, Zkn, Zk (RV64)          |
+| `int64_t _rv64_sha512sum1(int64_t rs1);`                      | `sha512sum1`  | Sum1 function for SHA2-512.               | Zknh, Zkn, Zk (RV64)          |
```
kito-cheng, line 196: I would suggest define `_rv64_sha512*` for both RV32 and RV64, and let compiler use `_rv64_sha512*h`  and `_rv64_sha512*l` to implement that for RV32, my assumption is those function are `sha512sig0` = `sha512sig0h` + `sha512sig0l`, and user should always use `sha512sig0h` + `sha512sig0l` on RV32, so just provide such interface would be easier use and porting.

riscv-c-api.md
```diff
@@ -114,7 +129,92 @@ For example:
...
+| `int64_t _rv64_aes64ks2(int64_t rs1, int64_t rs2);`           | `aes64ks2`    | AES key schedule, word mixing.            | Zkne, Zknd, Zkn, Zk (RV64)    |
+| `int32_t _rv32_brev8(int32_t rs1);`                           | `brev8`       | Reverse order of bits within each byte.   | Zbkb (RV32)                   |
+| `int64_t _rv64_brev8(int64_t rs1);`                           | `brev8`       | Reverse order of bits within each byte.   | Zbkb (RV64)                   |
+| `int32_t _rv32_clmul(int32_t rs1, int32_t rs2);`              | `clmul`       | Carry-less multiply (low 32 bits).        | Zbc, Zbkc (RV32)              |
+| `int64_t _rv64_clmul(int64_t rs1, int64_t rs2);`              | `clmul`       | Carry-less multiply (low 64 bits).        | Zbc, Zbkc (RV64)              |
+| `int64_t _rv64_clmulh(int64_t rs1, int64_t rs2);`             | `clmulh`      | Carry-less multiply (high 32 bits).       | Zbc, Zbkc (RV32)              |
+| `int32_t _rv32_clmulh(int32_t rs1, int32_t rs2);`             | `clmulh`      | Carry-less multiply (high 64 bits).       | Zbc, Zbkc (RV64)              |
```
kito-cheng, line 178: I guess those two prototype are put in reverse order?

riscv-c-api.md
```diff
@@ -114,7 +129,92 @@ For example:
...
+| `int64_t _rv64_aes64es(int64_t rs1, int64_t rs2);`            | `aes64es`     | AES final round encryption / RV64.        | Zkne, Zkn, Zk (RV64)          |
+| `int64_t _rv64_aes64esm(int64_t rs1, int64_t rs2);`           | `aes64esm`    | AES middle round encryption / RV64.       | Zkne, Zkn, Zk (RV64)          |
+| `int64_t _rv64_aes64im(int64_t rs1);`                         | `aes64im`     | AES Inverse MixColumns, key schedule.     | Zknd, Zkn, Zk (RV64)          |
+| `int64_t _rv64_aes64ks1i(int64_t rs1, int rnum);`             | `aes64ks1i`   | AES key schedule, round number.           | Zkne, Zknd, Zkn, Zk (RV64)    |
+| `int64_t _rv64_aes64ks2(int64_t rs1, int64_t rs2);`           | `aes64ks2`    | AES key schedule, word mixing.            | Zkne, Zknd, Zkn, Zk (RV64)    |
+| `int32_t _rv32_brev8(int32_t rs1);`                           | `brev8`       | Reverse order of bits within each byte.   | Zbkb (RV32)                   |
+| `int64_t _rv64_brev8(int64_t rs1);`                           | `brev8`       | Reverse order of bits within each byte.   | Zbkb (RV64)                   |
```
kito-cheng, line 174: Maybe provide `long _rv_brev8(long rs1)` only?

riscv-c-api.md
mjosaarinen: Agree that this might be elegant and would make the 32-bit and 64-bit implementations roughly equivalent. This is because the SHA2-512 loop mixes these functions with 64-bit additions. However, would break the assumption that intrinsics correspond to single instructions.

mjosaarinen: Added this as an option in https://github.com/mjosaarinen/riscv-c-api-doc/commit/db12510af20109ef17e7c9ba999d105c4fb9d350

riscv-c-api.md
mjosaarinen: It's a RISC-V convention to sign extend all 32-bit "W" intrinsics; even things like 32-bit rotation on RV64. The architecture review committee asked the SHA256 instructions to be like that too, for consistency, so they are like that. Everything is sign-exteded -- Indeed this trips a lot programmers. Also, what's precisely why the (signed) "long" type was used in bitmanip (they were all sign extended at least since 2019 -- so not as you recall). If one uses unsigned types then there is a risk of automatic generation of instructions that clear the upper half.

kito-cheng: I am not sure I've got the point about sign extension? long type in kind of more `intXLEN_t`, so there is no sign-extension issue on RV64 I think?

riscv-c-api.md
```diff
@@ -114,7 +129,96 @@ For example:
...
+| `int64_t _rv64_sha512sum1(int64_t rs1);`                      | `sha512sum1`  | Sum1 function for SHA2-512.               | Zknh, Zkn, Zk (RV64 [RV32])   |
+| `long _rv_sm3p0(long rs1);`                                   | `sm3p0`       | P0 function for SM3 hash.                 | Zksh, Zks (RV32,RV64)         |
+| `long _rv_sm3p1(long rs1);`                                   | `sm3p1`       | P1 function for SM3 hash.                 | Zksh, Zks (RV32,RV64)         |
+| `long _rv_sm4ed(int32_t rs1, int32_t rs2, int bs);`           | `sm4ed`       | Accelerate SM4 cipher encrypt/decrypt.    | Zksed, Zks (RV32,RV64)        |
+| `long _rv_sm4ks(int32_t rs1, int32_t rs2, int bs);`           | `sm4ks`       | Accelerate SM4 cipher key schedule.       | Zksed, Zks (RV32,RV64)        |
+| `int32_t _rv32_unzip(int32_t rs1);`                           | `unzip`       | Odd/even bits into upper/lower halves.    | Zbkb (RV32)                   |
+| `int32_t _rv32_xperm4(int32_t rs1, int32_t rs2);`             | `xperm4`      | Nibble-wise lookup of indicies.           | Zbkx (RV32)                   |
```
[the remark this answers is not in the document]
mjosaarinen: it's actually in the spec using that word too, unfortunately. anyway, thanks! https://github.com/mjosaarinen/riscv-c-api-doc/commit/c6a30248c3def14bce986d5109a70a7d5428b5fb

riscv-c-api.md
kito-cheng: I guess I only thinking about the interface itself is equivalence and not consider the semantic/function is different on RV32/RV64, different name/interface type should be more clear to let user aware that, I am OK with current design.

riscv-c-api.md
```diff
@@ -114,7 +129,96 @@ For example:
...
+| `int64_t _rv64_aes64ks1i(int64_t rs1, int rnum);`             | `aes64ks1i`   | AES key schedule, round number.           | Zkne, Zknd, Zkn, Zk (RV64)    |
+| `int64_t _rv64_aes64ks2(int64_t rs1, int64_t rs2);`           | `aes64ks2`    | AES key schedule, word mixing.            | Zkne, Zknd, Zkn, Zk (RV64)    |
+| `int32_t _rv32_brev8(int32_t rs1);`                           | `brev8`       | Reverse order of bits within each byte.   | Zbkb (RV32)                   |
+| `int64_t _rv64_brev8(int64_t rs1);`                           | `brev8`       | Reverse order of bits within each byte.   | Zbkb (RV64)                   |
+| `int32_t _rv32_clmul(int32_t rs1, int32_t rs2);`              | `clmul`       | Carry-less multiply (low 32 bits).        | Zbc, Zbkc (RV32)              |
+| `int64_t _rv64_clmul(int64_t rs1, int64_t rs2);`              | `clmul`       | Carry-less multiply (low 64 bits).        | Zbc, Zbkc (RV64)              |
+| `int32_t _rv32_clmulh(int32_t rs1, int32_t rs2);`             | `clmulh`      | Carry-less multiply (high 64 bits).       | Zbc, Zbkc (RV64)              |
```
topperc, line 181: Are the descriptions for clmulh swapped between 32 and 64 bit.

riscv-c-api.md
```diff
@@ -114,7 +129,96 @@ For example:
...
+| `int64_t _rv64_sha512sig1(int64_t rs1);`                      | `sha512sig1`  | Sigma1 function for SHA2-512.             | Zknh, Zkn, Zk (RV64 [RV32])   |
+| `int64_t _rv64_sha512sum0(int64_t rs1);`                      | `sha512sum0`  | Sum0 function for SHA2-512.               | Zknh, Zkn, Zk (RV64 [RV32])   |
+| `int64_t _rv64_sha512sum1(int64_t rs1);`                      | `sha512sum1`  | Sum1 function for SHA2-512.               | Zknh, Zkn, Zk (RV64 [RV32])   |
+| `long _rv_sm3p0(long rs1);`                                   | `sm3p0`       | P0 function for SM3 hash.                 | Zksh, Zks (RV32,RV64)         |
+| `long _rv_sm3p1(long rs1);`                                   | `sm3p1`       | P1 function for SM3 hash.                 | Zksh, Zks (RV32,RV64)         |
+| `long _rv_sm4ed(int32_t rs1, int32_t rs2, int bs);`           | `sm4ed`       | Accelerate SM4 cipher encrypt/decrypt.    | Zksed, Zks (RV32,RV64)        |
+| `long _rv_sm4ks(int32_t rs1, int32_t rs2, int bs);`           | `sm4ks`       | Accelerate SM4 cipher key schedule.       | Zksed, Zks (RV32,RV64)        |
```
topperc, line 204: Is the return type of long important here or could it be int32_t? I know the result returns a sign extended value on RV64. _rv32_ror doesn't return long for 64-bits. Or is that because _rv32_ror came from bitmanip and if you defined it from scratch it would return long?

mjosaarinen: It was defined this way in bitmanip's `rvintrin.h`. I agree that 32-bit ror and rol could be defined with "long" for consistency, as it is available in both RV32 and RV64. On the other hand I see no great danger in keeping it this way.

aswaterman: Instructions that are guaranteed to return a sign-extended 32-bit result should definitely return `int32_t` or equivalent.  This tightens up the API a little bit, and it'll probably obviate some stray `sext.w` instructions in some cases under some compilers.

ZenithalHourlyRate:
> | __riscv_zknh            | Arch Version | `Zknh` extension is available.                                    |

About the Arch Version, I found that when using the latest [riscv-collab/riscv-gnu-toolchain](https://github.com/riscv-collab/riscv-gnu-toolchain), with `-march="rv64gc_zknh"`, the value of `__riscv_zknh` is 0, and with `-march="rv64gc_zknh1"`, the value is then 1000000 (correct according to the spec).

So what is the meaning of `__riscv_zknh` being zero, is zknh available to the software? This affects how software src detects whether the compiler supports zknh extension (e.g. https://github.com/openssl/openssl/pull/16638#issuecomment-1121281173)

PR #24, opened 2022-01-18
Use Zb[abcs] instead of B/Bitmanip

cmuellner:
We don't have a single B/Bitmanip extension, but several extensions
Zb[abcs], which can be available independent of each other.
Let's fix that in this document.
Given that there was never a ratified Bitmanip extensions,
I don't think that this change causes any real-world problems.

@kito-cheng @jim-wilson @asb

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -56,7 +56,6 @@ which is compute by following formula:
 
 For example:
 - F-extension v2.2 will define `__riscv_f` as `2002000`.
-- B-extension v0.92 will define `__riscv_b` as `92000`.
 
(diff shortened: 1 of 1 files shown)


riscv-c-api.md
```diff
@@ -70,6 +69,10 @@ For example:
 | __riscv_c               | Arch Version | `C` extension is available.   |
 | __riscv_b               | Arch Version | `B` extension is available.   |
```
kito-cheng, line 70: Could you also remove this?

kito-cheng:
Otherwise, LGTM, thanks!

asb:
This LGTM.

PR #26, opened 2022-02-09
Add pre-defined macro __riscv_p for P extension

tclin914:


kito-cheng:
I believe this change is straightforward and no controversy since https://github.com/riscv/riscv-p-spec is there for long time.

cmuellner:
I'm not a big fan of merging this change, because the P extension is not frozen.

kito-cheng:
Oh, apology for the rush merge, but I think that's still useful for implementing compiler support for those non-ratified extensions, what do you think if we add few more word to mark that is non-ratified / non-frozen?

cmuellner:
I would prefer to keep changes for unfrozen extensions as PRs and tag them with "unfrozen extension".
Once the extension is frozen, we can merge the change. However, I don't want to overtake the extension ratification process.

My plan is to create a proposal for a merge policy, that we can discuss.

kito-cheng:
> My plan is to create a proposal for a merge policy, that we can discuss.

OK, thanks!

tclin914:
Hi @kito-cheng,

Do you plan to revert these changes?
I will create a new PR the same as this. It seems merged PR unable to be reopened.

PR #27, opened 2022-02-14
Add architecture extension test macro for all ratified vector extensions

kito-cheng:
Add zve* and zvl*b extensions.

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -112,23 +112,40 @@ which is compute by following formula:
 For example:
 - F-extension v2.2 will define `__riscv_f` as `2002000`.
 
-| Name                    | Value        | When defined                  |
-| ----------------------- | ------------ | ----------------------------- |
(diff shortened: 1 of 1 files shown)


kito-cheng:
@nick-knight @asb @frasercrmck @topperc @rdolbeau

asb:
Thanks for taking a stab at updating this KIto. Other ratified extensions aren't represented of course, like zicbop, ziboz, zicbom, zfhmin, the scalar crypto extensions. I'm wondering if it's more trouble than it's worth to keep this table as an exhaustive list, and it might be better to just clearly explain that supported extensions have __riscv_foo defined, and give some examples?

kito-cheng:
@asb Sounds like that's right way to do, here is the PR: https://github.com/riscv-non-isa/riscv-c-api-doc/pull/28, thanks!

kito-cheng:
And I'll reshape this PR to contain all ratified extensions if possible

PR #28, opened 2022-02-14
Explicitly write down the rule of architecture extension test macros

kito-cheng:
We didn't write the rule explicitly, so people might not expect those marco
to be available if it's not appear on the table.

So let document that, and add more note to say people can expect that
even not list on the table.

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -102,6 +102,14 @@ availability and version for certain extension, however not all compilers are
 supported, so you should check `__riscv_arch_test` to make sure this compiler
 is supporting those preprocessor definitions.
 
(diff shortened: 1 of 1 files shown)


asb:
I've left a couple of minor phrasing suggestions.

One alternate approach would be to cut down the table and just list a few examples, on the basis that keeping it up to date is extra work that might not have much value for end users.

kito-cheng:
@asb Seems like those suggestion are missing by some accident? Did you mind send again?

asb:
> @asb Seems like those suggestion are missing by some accident? Did you mind reply those suggestion again?

Ah, they were still pending and unsubmitted. Fixed now.

PR #29, opened 2022-04-26
Zawrs: Define extension availability macro

cmuellner:
This patch defines an availability macro for the Zawrs extension.

Note, that Zawrs is not frozen or ratified yet. I.e. this is an RFC for the time being.


kito-cheng (approved):
Oh, just found this still their, anyway let merge first.

PR #31, opened 2022-11-21
Intrinsics: Define intrinsics naming guidelines

cmuellner:
Let's establish a naming guideline for intrinsics.

Contrary to the previous attempt in #25, this PR does not try to make a guideline that covers all existing formats but defines a single new format.

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -172,5 +172,44 @@ For example:
 
 ## Intrinsic Functions
 
-Do we really have none of these?  I can't figure out
-`gcc/gcc/config/riscv/riscv-builtins.c`...
+Intrinsic functions (or intrinsics or built-ins) are expanded into instruction sequences by compilers.
+They typically provide access to functionality that is otherwise not synthesizable by compilers.
+Some intrinsics expand to different code sequences depending on the available instructions from the enabled ISA extensions.
+
+Compilers typically come with their own architecture-independent intrinsics (e.g. synchronization primitives, byte-swap, etc.).
+The RISC-V compiler backend can define additional target-specific intrinsics.
+Providing functionality via architecture-independent intrinsics is the preferred method, as it improves code portability.
+
+Some intrinsics are only available if a particular header file is included.
+RISC-V header files that enable intrinsics require the prefix `riscv_` (e.g. `riscv_vector.h` or `riscv_crypto.h`).
+
+RISC-V specific intrinsics use the common prefix "__riscv_" to avoid namespace collisions.
+
+The intrinsic name describes the functional behaviour of the function.
+In case the functionality can be expressed with a single instruction, the instruction's name (any '.' replaced by '_') is the preferred choice.
+Note, that intrinsics that are restricted to RISC-V vendor extensions need to include the vendor prefix (as documented in the RISC-V toolchain conventions).
+
+If intrinsics are available for multiple data types, then function overloading is preferred over multiple type-specific functions.
+If an intrinsic function is has parameters or return values that reference registers with XLEN bits, then the data type `long` should be used.
+In case a function is only available for one data type and this type cannot be derived from the function's name, then the type should be appended to the function name, delimited by a '_' character.
+Typical type postfixes are "32" (32-bit), "i32" (signed 32-bit), "i8m4" (vector register group consisting of 4 signed 8-bit vector registers).
+
+RISC-V intrinsics follow the following naming rule:
+
+```
+INTRINSIC ::= PREFIX NAME [ '_' TYPE ]
+PREFIX ::= "__riscv_"
+NAME ::= Name of the intrinsic function.
+TYPE ::= Optional type postfix.
+```
+
+RISC-V intrinsics examples:
+
+```
+type __riscv_orc_b (type rs); // orc.b rd, rs
+
+long __riscv_clmul (long a, long b); // clmul rd, rs1, rs2
+
+#include <riscv_vector.h> // make RISC-V vector intrinsics available
+vint8m1_t __riscv_vadd_vv_i8m1(vint8m1_t vs2, vint8m1_t vs1, size_t vl); // vadd.vv vd, vs2, vs1
+```


cmuellner:
CC @kito-cheng @asb

cmuellner:
CC @eopXD

riscv-c-api.md
```diff
@@ -172,5 +172,36 @@ For example:
...
+
+```
+RISC-V intrinsics examples:
+
+orc.b rd, rs:
+type __rv_orc_b(type rs);
+int32_t __rv_orc_b_32(int32_t rs);
+
+vadd.vv vd, vs2, vs1:
+vint8m1_t __rv_vadd_vv_i8m1(vint8m1_t vs2, vint8m1_t vs1, size_t vl);
```
eopXD, line 206: Probably missing an option naming rule on "type abbreviation" (e.g. the `i` here)

cmuellner:
@ptomsich mentioned in today's SIG Toolchain call that he prefers `__riscv_` as prefix over `__rv_`. There were no objections, so I'll update the PR accordingly.

rjiejie:
Shall we need to consider about customized or vendor's intrinsic name ?
like __riscv_<vendor>_ ?
e.g. __riscv_thead_

riscv-c-api.md
```diff
@@ -172,5 +172,36 @@ For example:
 
 ## Intrinsic Functions
 
-Do we really have none of these?  I can't figure out
-`gcc/gcc/config/riscv/riscv-builtins.c`...
+Intrinsic functions (or intrinsics or built-ins) are functions that are expanded into instruction sequences by compilers.
+They typically provide access to functionality that is otherwise not synthesizable by compilers.
+Some intrinsics expand to different code sequenzes, depending on the available instructions from the enabled ISA extensions.
```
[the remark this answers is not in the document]
cmuellner: Thanks!

riscv-c-api.md
cmuellner: Will add. Thanks!

cmuellner:
I've updated the PR according to the feedback so far.

Regarding the vendor-specific intrinsics, I have no opinion yet. But I think the proposal sounds reasonable. Thanks!

cmuellner:
Regarding the vendor instructions, I don't see why we need a vendor prefix because the instructions are already prefixed (e.g. `th.mula` -> `__riscv_th_mula`). Do you think we need more?

cmuellner:
> `__riscv_th` or `__riscv_thead` are both valid on current scheme I think.

I think that all vendor-specific intrinsics would only be covering vendor-instructions. In this case, the instruction's name is the preferred choice for the name. Therefore `__riscv_th` would be used for T-Head-specific intrinsics.

rjiejie:
@cmuellner @kito-cheng @eopXD Hello ? nobody care about Vendor's input ?

rjiejie:
> Regarding the vendor instructions, I don't see why we need a vendor prefix because the instructions are already prefixed (e.g. `th.mula` -> `__riscv_th_mula`). Do you think we need more?

Looks like it's a choice although the prefix 'th' have no obvious meaning about vendor's info in user side :)

cmuellner:
> > Regarding the vendor instructions, I don't see why we need a vendor prefix because the instructions are already prefixed (e.g. `th.mula` -> `__riscv_th_mula`). Do you think we need more?
> 
> Looks like it's a choice although the prefix 'th' have no obvious meaning about vendor's info in user side :)

The vendor prefixes are required for vendor instructions.
Therefore they are standardized in the riscv-toolchain-conventions: https://github.com/riscv-non-isa/riscv-toolchain-conventions/blob/master/README.mkd#list-of-vendor-prefixes

E.g. in Binutils all instructions of the T-Head vendor extensions have the prefix "th" (e.g. "th.ldd" which is part of XTheadMemPair).

I'll add the following sentence:
```
Note, that intrinsics that are restricted to RISC-V vendor extensions need to include the vendor-prefix (as documented in the RISC-V toolchain conventions).
```

riscv-c-api.md
```diff
@@ -172,5 +172,48 @@ For example:
...
+Intrinsic functions (or intrinsics or built-ins) are expanded into instruction sequences by compilers.
+They typically provide access to functionality that is otherwise not synthesizable by compilers.
+Some intrinsics expand to different code sequences depending on the available instructions from the enabled ISA extensions.
+
+Compilers typically come with their own architecture-independent intrinsics (e.g. synchronization primitives, byte-swap, etc.).
+The RISC-V compiler backend can define additional target-specific intrinsics.
+Providing functionality via architecture-independent intrinsics is the preferred method, as it improves code portability.
+
+Some intrinsics are only available if a particular header file is included.
+RISC-V header files that enable intrinsics require the prefix `riscv_` (e.g. `riscv_vector.h` or `riscv_crypto.h`).
```
fanghuaqi, line 184: Is there any rules about the header file naming if header file required?

kito-cheng (approved):
Giving my blessing of this PR, hope this could push this forward :)

cmuellner:
Thanks for the approval!
I'll merge this later today.

riscv-c-api.md
cmuellner: Short answer: no rules beyond the prefix `riscv_`.

The reason is that this guide cannot foresee which grouping makes sense.
E.g. vector crypto instructions might be gated by `riscv_vector.h`, `riscv_vector_crypto.h`, or `riscv_crypto.h`. All of them would be valid candidates (maybe even multiple options), so this will have to be decided on a case-by-case basis.

cmuellner:
I got the feedback that the CMO examples are confusing. Therefore I dropped them.

cmuellner:
This was discussed (again) in today's SIG Toolchain call, and there were no objections to this proposal.

riscv-c-api.md
fanghuaqi: Got it, and it will be good to have same filename naming rule for different toolchains.

nick-knight:
I'm hesitant to say anything, because adding the `__riscv_` prefix to our vector intrinsics across the codebase has been a huge productivity killer this week. And it makes the already-lengthy vector intrinsics even more annoying to type. But I just want to confirm that you aren't also planning to require some kind of prefix for the RVV intrinsics types (`vint8m8_t`, etc.).

Firdous24:
Which version/release/branch of riscv-gnu-toolchain supports the prefix "__riscv_" in rvv intrinsics?

Firdous24:
Which version/release/branch of riscv-gnu-toolchain supports the prefix "__riscv_" in RVV intrinsic? As I have cloned the latest version of toolchain but it doesn't support the __riscv_ prefix.

PR #33, opened 2023-03-27
Add operand constraint for inline asm statement.

kito-cheng:
Add miminal common operand constraint between LLVM and GCC.

Plan to extend this like vector register constraint in future.

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -213,3 +213,66 @@ long __riscv_clmul (long a, long b); // clmul rd, rs1, rs2
 #include <riscv_vector.h> // make RISC-V vector intrinsics available
 vint8m1_t __riscv_vadd_vv_i8m1(vint8m1_t vs2, vint8m1_t vs1, size_t vl); // vadd.vv vd, vs2, vs1
 ```
+
+## Constraints on Operands of Inline Assembly Statements
+
+This section lists operand constraints that can be used with inline assembly
+statements, including both RISC-V specific and common operand constraints.
+
+
+| Constraint         |                                    | Note        |
+| ------------------ | ---------------------------------- | ----------- |
+| m                  | An address that is held in a general-purpose register with offset.      |      |
+| A                  | An address that is held in a general-purpose register.      |      |
+| r                  | General purpose register                      |      |
+| f                  | Floating-point register                       |      |
+| i                  | Immediate integer operand                     |      |
+| I                  | 12-bit signed immediate integer operand      |      |
+| K                  | 5-bit unsigned immediate integer operand     |      |
+| J                  | Zero integer immediate operand                |      |
+
+NOTE: Immediate value must be a compile-time constant.
+
+### The Difference Between `m` and `A` Constraints
+
+The difference between `m` and `A` is whether the operand can have an offset;
+some instructions in RISC-V do not allow an offset for the address operand,
+such as atomic or vector load/store instructions.
+
+The following example demonstrates the difference; it is trying
+to load value from `foo[10]` and using `m` and `A` to pass that address.
+
+```c
+int *foo;
+void bar() {
+ int x;
+ __asm__ volatile ("lw %0, %1" : "=r"(x) : "m" (foo[10]));
+ __asm__ volatile ("lw %0, %1" : "=r"(x) : "A" (foo[10]));
+}
+```
+
+Then we compile with GCC with `-O` option:
+
+```shell
+$ riscv64-unknown-elf-gcc x.c -o - -O -S
+...
+bar:
+       lui    a5,%hi(foo)
+       ld     a5,%lo(foo)(a5)
+ #APP
+# 4 "x.c" 1
+       lw a4, 40(a5)
+# 0 "" 2
+ #NO_APP
+       addi   a5,a5,40
+ #APP
+# 5 "x.c" 1
+       lw a5, 0(a5)
+# 0 "" 2
+ #NO_APP
+       ret
+
+```
+
+The compiler uses an immediate offset of 40 for the `m` constraint, but for the
+`A` constraint uses an extra addi instruction instead.


riscv-c-api.md
```diff
@@ -174,3 +174,72 @@ For example:
...
 Do we really have none of these?  I can't figure out
 `gcc/gcc/config/riscv/riscv-builtins.c`...
+
+## Constraints for Operands of Inline Assembly Statements
+
+Mainstream compilers like GCC and Clang/LLVM support writing inline assembly
+language in C and C++ programs by `__asm__` or `asm` statement; RISC-V follows
+ the common* way of using assembly language.
+
+This section lists operands constraint that can be used with inline assembly
```
enh-google, line 184: operand constraints

riscv-c-api.md
```diff
@@ -174,3 +174,72 @@ For example:
 
 Do we really have none of these?  I can't figure out
 `gcc/gcc/config/riscv/riscv-builtins.c`...
+
+## Constraints for Operands of Inline Assembly Statements
```
enh-google, line 178: Constraints _on_ Operands...

riscv-c-api.md
```diff
@@ -174,3 +174,72 @@ For example:
...
+
+Mainstream compilers like GCC and Clang/LLVM support writing inline assembly
+language in C and C++ programs by `__asm__` or `asm` statement; RISC-V follows
+ the common* way of using assembly language.
```
enh-google, line 182: drop everything after the `;`? there's a stray `*`, the text is not quite grammatical, and it's not obvious to me what you're trying to say anyway?

(tbh, i'd have just dropped this entire paragraph. anyone who doesn't already know what's in this paragraph isn't going to be able to cope any better with following paragraphs for having read this :-) )

riscv-c-api.md
```diff
@@ -174,3 +174,72 @@ For example:
...
+
+
+| Constraint         |                                    | Note        |
+| ------------------ | ---------------------------------- | ----------- |
+| m                  | An address that is held in a general-purpose register with offset.      |      |
+| A                  | An address that is held in a general-purpose register.      |      |
+| r                  | General purpose register                      |      |
+| f                  | Floating-point register                       |      |
+| i                  | Immediate integer operand                     |      |
+| I                  | 12-bits signed immediate integer operand      |      |
```
enh-google, line 195: bit

riscv-c-api.md
```diff
@@ -174,3 +174,72 @@ For example:
...
+| K                  | 5-bits unsigned immediate integer operand     |      |
+| J                  | Zero integer immediate operand                |      |
+
+NOTE: Immediate value must be compile-time constant.
+
+### The Difference Between `m` and `A` Constraint
```
enh-google, line 201: Constraints

riscv-c-api.md
```diff
@@ -174,3 +174,72 @@ For example:
...
+
+The difference between `m` and `A` is whether the operand can have offset;
```
enh-google, line 203: have an

riscv-c-api.md
```diff
@@ -174,3 +174,72 @@ For example:
...
+some instructions in RISC-V do not allow offset for the address operand, like
```
enh-google, line 204: allow an

enh-google, line 204: s/like/such as/

riscv-c-api.md
```diff
@@ -174,3 +174,72 @@ For example:
...
+atomic or vector load/store instructions.
+
+Give an example to demonstrate the difference; the following program is trying
```
enh-google, line 207: The following example demonstrates the difference; it is trying

riscv-c-api.md
```diff
@@ -174,3 +174,72 @@ For example:
...
+```c
+int *foo;
+void bar() {
+ int x;
+ __asm__ volatile ("lw %0, %1" : "=r"(x) : "m" (foo[10]));
+ __asm__ volatile ("lw %0, %1" : "=r"(x) : "A" (foo[10]));
+}
+```
+
+Then we compile with GCC with O3 option:
```
enh-google, line 219: the example says `-O`, not `-O3`

riscv-c-api.md
```diff
@@ -174,3 +174,72 @@ For example:
...
+ #APP
+# 5 "x.c" 1
+       lw a5, 0(a5)
+# 0 "" 2
+ #NO_APP
+       ret
+
+```
+
+Then we can observe that the compiler will generate the form of the base register
```
enh-google, line 242: The compiler uses an immediate offset of 40 for the `m` constraint, but for the `A` constraint uses an extra `addi` instruction instead.

riscv-c-api.md
kito-cheng: Dropped

kito-cheng:
@enh-google thanks! updated per your comment, I guess I should pay more to [grammarly](https://www.grammarly.com/) ...:P

enh-google (approved):

riscv-c-api.md
```diff
@@ -213,3 +213,66 @@ long __riscv_clmul (long a, long b); // clmul rd, rs1, rs2
...
+
+Then we compile with GCC with `-O` option:
+
+```shell
+$ riscv64-unknown-elf-gcc x.c -o - -O -S
+...
+bar:
+       lui    a5,%hi(foo)
+       ld     a5,%lo(foo)(a5)
+ #APP
```
cmuellner, line 262: What does "APP" stand for?

cmuellner (approved):
LGTM

riscv-c-api.md
MaskRay: APP is comment markers from Clang's inline assembly handling code...

MaskRay (approved):

riscv-c-api.md
kito-cheng: Seems like it has some function on gnu Assembler Used to turning preprocessing on and off:

https://sourceware.org/binutils/docs/as/Preprocessing.html
```
If the first line of an input file is #NO_APP or if you use the ‘-f’ option, whitespace and comments are not removed from the input file. Within an input file, you can ask for whitespace and comment removal in specific portions of the file by putting a line that says #APP before the text that may contain whitespace or comments, and putting a line that says #NO_APP after this text. This feature is mainly intended to support asm statements in compilers whose output is otherwise free of comments and whitespace.
```

And from GCC doc: https://gcc.gnu.org/onlinedocs/gccint/File-Framework.html
```
Macro: ASM_APP_ON
A C string constant for text to be output before each asm statement or group of consecutive ones. Normally this is "#APP", which is a comment that has no effect on most assemblers but tells the GNU assembler that it must check the lines that follow for all valid assembler constructs.

Macro: ASM_APP_OFF
A C string constant for text to be output after each asm statement or group of consecutive ones. Normally this is "#NO_APP", which tells the GNU assembler to resume making the time-saving assumptions that are valid for ordinary compiler output.
```


But I guess that is just a marker for clang/llvm.

BTW, that's the first time I figure out what APP/NO_APP means.

PR #34, opened 2023-03-31
Add more description for those function attributes

kito-cheng:
We have few RISC-V specific function attributes like naked and interrupt, they are implmented on both LLVM and GCC for a while.

But we don't have well described here.

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -164,11 +164,32 @@ For example:
 
 ## Function Attributes
 
-* `__attribute__((naked))`
-* `__attribute__((interrupt))`
-* `__attribute__((interrupt("user")))`
-* `__attribute__((interrupt("supervisor")))`
-* `__attribute__((interrupt("machine")))`
+### `__attribute__((naked))`
+
+The compiler won't generate the prologue/epilogue for those functions with
+`naked` attributes. This attribute is usually used when you want to write a
+function with an inline assembly body.
+
+This attribute is incompatible with the `interrupt` attribute.
+
+NOTE: Be aware that compilers might have further restrictions on naked
+functions. Please consult your compiler's manual for more information.
+
+### `__attribute__((interrupt))`, `__attribute__((interrupt("user")))`, `__attribute__((interrupt("supervisor")))` `__attribute__((interrupt("machine")))`
+
+The interrupt attribute specifies that a function is an interrupt handler.
+The compiler will save/restore all used registers in the prologue/epilogue
+regardless of the ABI, all used registers including floating point
+register/vector register if `F` extension/vector extension is enabled.
+
+The interrupt attribute can have an optional parameter to specify the mode.
+The possible values are `user`, `supervisor`, or `machine`.
+The default value `machine` is used, if the mode is not specified.
+
+The function can specify only one mode; the compiler should raise an error if a
+function declares more than one mode or an undefined mode.
+
+This attribute is incompatible with the `naked` attribute.
 
 ## Intrinsic Functions
 


kito-cheng:
cc @asb @luismarques

riscv-c-api.md
```diff
@@ -164,11 +164,26 @@ For example:
...
+The compiler won't generate the prologue/epilogue for those functions with
+`naked` attributes, that usually used when you want to write a function with
+an inline assembly body.
+
+This attribute is incompatible with `interrupt` attribute.
+
+### `__attribute__((interrupt))`, `__attribute__((interrupt("user")))`, `__attribute__((interrupt("supervisor")))` `__attribute__((interrupt("machine")))`
+
+`interrupt` used to specified a function is an interrupt handler. The compiler
+will save/restore all used register at prologue/epilogue regardless of the ABI.
```
cmuellner, line 178: What is "all used register"? All GPRs? What about float/vector registers?

riscv-c-api.md
```diff
@@ -164,11 +164,26 @@ For example:
...
 
-* `__attribute__((naked))`
-* `__attribute__((interrupt))`
-* `__attribute__((interrupt("user")))`
-* `__attribute__((interrupt("supervisor")))`
-* `__attribute__((interrupt("machine")))`
+### `__attribute__((naked))`
+
+The compiler won't generate the prologue/epilogue for those functions with
+`naked` attributes, that usually used when you want to write a function with
```
cmuellner, line 170: Make 2 sentences: ...attributes. This attribute is usually used...

riscv-c-api.md
```diff
@@ -164,11 +164,26 @@ For example:
...
+an inline assembly body.
+
+This attribute is incompatible with `interrupt` attribute.
```
cmuellner, line 173: ...with the `interrupt`...

riscv-c-api.md
```diff
@@ -164,11 +164,26 @@ For example:
...
+
+### `__attribute__((interrupt))`, `__attribute__((interrupt("user")))`, `__attribute__((interrupt("supervisor")))` `__attribute__((interrupt("machine")))`
+
+`interrupt` used to specified a function is an interrupt handler. The compiler
```
cmuellner, line 177: The `interrupt` attribute specifies that a function is an interrupt handler.

riscv-c-api.md
cmuellner, line 178: all used registers in the prologue/epilogue

riscv-c-api.md
```diff
@@ -164,11 +164,26 @@ For example:
...
+will save/restore all used register at prologue/epilogue regardless of the ABI.
+
+`interrupt` can have an optional parameter for specify the mode, it can be
```
cmuellner, line 180: The `interrupt` attribute can have an optional parameter to specify the mode.
The possible values are `user`, `supervisor`, or `machine`.
The default value `machine` is used, if the mode is not specified.

riscv-c-api.md
```diff
@@ -164,11 +164,26 @@ For example:
...
+`user`, `supervisor` or `machine`, default is `machine` if not specified.
+
+The function can specify only one mode; the compiler should raise an error if a
+function declares more than one mode or conflicted mode.
```
cmuellner, line 184: What is "conflicted" mode?
Probably: "...one mode or an undefined mode."

cmuellner (commented):
LGTM, just few suggestions to improve the spelling.

riscv-c-api.md
```diff
@@ -164,11 +164,26 @@ For example:
...
-* `__attribute__((naked))`
-* `__attribute__((interrupt))`
-* `__attribute__((interrupt("user")))`
-* `__attribute__((interrupt("supervisor")))`
-* `__attribute__((interrupt("machine")))`
+### `__attribute__((naked))`
+
+The compiler won't generate the prologue/epilogue for those functions with
+`naked` attributes, that usually used when you want to write a function with
+an inline assembly body.
```
asb, line 171: We're getting to the point where we risk just reproducing documentation that should really belong in the compilers docs, but a key difference here is that non-asm statements are not allowed in naked functions in Clang. Unsure if it's worth mentioning this or not.

cmuellner: We should not duplicate compiler-specifics here. They might change, and we end up with outdated information here.

Still, the hint is helpful. What about a more generic note like "Note: Be aware that compilers might have further restrictions on naked functions. Please consult your compiler's manual for more information."

kito-cheng: Actually I am prepare to propose `target` attribute, but I found function attribute section already there but almost empty body, so I try to fill this empty before I post `target` attribute part.

For `naked` and `interrupt`, that might seem like reproducing doc. but `target` won't be the same case I believe :)

kito-cheng:
Changes:

- Apply @cmuellner's comment

kito-cheng:
ping

riscv-c-api.md
```diff
@@ -164,11 +164,32 @@ For example:
...
+
+This attribute is incompatible with the `interrupt` attribute.
+
+NOTE: Be aware that compilers might have further restrictions on naked
+functions. Please consult your compiler's manual for more information.
+
+### `__attribute__((interrupt))`, `__attribute__((interrupt("user")))`, `__attribute__((interrupt("supervisor")))` `__attribute__((interrupt("machine")))`
+
+The interrupt attribute specifies that a function is an interrupt handler.
+The compiler will save/restore all used register in the prologue/epilogue
```
cmuellner, line 181: "all used registers"

cmuellner (approved):
LGTM

PR #35, opened 2023-04-02
Adding target attribute for function.

kito-cheng:
After adding mapping symbol and `.option arch`, finally we have
all necessary components of function multi-versioning.

This proposal has add a new funciton attribte: `target` attribte for compile a
funciton with specify extension, pipeline model or cost model, could be used as
simple`*` function multi-versioning.

`*`: Simple is compare to `target_clones` in x86 and `target_version` in
AArch64, and also require user to dispatch that right.

It similar to AArch64's target attribte: https://gcc.gnu.org/onlinedocs/gcc/AArch64-Function-Attributes.html

This attribte won't change the function name mangling, and also require
do not update Tag_RISCV_arch attribte.

Ref:
- https://maskray.me/blog/2023-02-05-function-multi-versioning

---

NOTE: This PR will not be merged until we have PoC *AND* GNU/LLVM community reach consensus 
NOTE: We don't have PoC for either GCC and clang yet,

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -203,6 +203,65 @@ function declares more than one mode or an undefined mode.
 
 This attribute is incompatible with the `naked` attribute.
 
+### `__attribute__((target("<ATTR-STRING>")))`
+
+`target` attribute used for enable or disable a set of features or extensions
+for the fucntion.
+
+Function with target attribute should not affect the file scope build
+attributes, i.g. File complied with -march=rv64ima and a function declares with
+`__attribute__((target("arch=+zbb")))`, the `Tag_RISCV_arch` build attribute
+should be `rv64ima` rather than `rv64ima_zbb`.
+
+The compiler may emit mapping symbol at the beginning of the function with
+target attribute if the function use different set of ISA extension.
+
+`<ATTR-STRING>` can be specify following target attributes:
+
+- `arch=`: Adding extra extensions, omit certain extension, or specific
+           extension setting for the function.
+- `tune=`: Specify the pipeline mode and cost model with specific
+           microarchitecture or core for the function.
+- `cpu=`: Specify the pipeline mode, cost model and extension setting for the
+          function.
+
+The syntax of `<ATTR-STRING>` describes below:
+
+```
+ATTR-STRING := ATTR ';' ATTR
+             | ATTR
+
+ATTR        := ARCH-ATTR
+             | CPU-ATTR
+             | TUNE-ATTR
+
+ARCH-ATTR   := 'arch=' EXTENSIONS-OR-FULLARCH
+
+EXTENSIONS-OR-FULLARCH := <EXTENSIONS>
+                        | <FULLARCHSTR>
+
+EXTENSIONS             := <EXTENSION> ',' <EXTENSIONS>
+                        | <EXTENSION>
+
+FULLARCHSTR            := <full-arch-string>
+
+EXTENSION              := <OP> <EXTENSION-NAME> <VERSION>
+
+OP                     := '+'
+                        | '-'
+
+VERSION                := [0-9]+ 'p' [0-9]+
+                        | [1-9][0-9]*
+                        |
+
+EXTENSION-NAME         := Naming rule is defined in RISC-V ISA manual
+
+CPU-ATTR    := 'cpu=' <valid-cpu-name>
+TUNE-ATTR   := 'tune=' <valid-tune-name>
+```
+
+NOTE: The `target` attribute won't effect function name mangling for both C and C++.
+
 ## Intrinsic Functions
 
 Intrinsic functions (or intrinsics or built-ins) are expanded into instruction sequences by compilers.
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -248,9 +248,14 @@ equivalent to `arch=rv64g;tune=sifive-7-series`, and
 `cpu=sifive-u74;tune=sifive-5-series` is equivalent to
 `arch=rv64gc;tune=sifive-5-series`.
 
-If the same type of attribute is specified more than once, only the latest one
-takes effect. For example, `arch=+zbb;arch=+zba` will be equivalent to
-`arch=+zba`.
+The compiler should emit error if the same type of attribute is specified more
+than once. For example, `arch=+zbb;arch=+zba`, compiler should emit error
+because `arch` has specified twice.
+
+The compiler should emit errof if target attribute has specified more than once.
+For example,
+`__attribute__((target("arch=+v"))) __attribute__((target("arch=+zbb"))) int foo(int a)`
+, compiler should emit error because target attribute has specified twice.
 
 The interactions between the attribute and the command-line option are
 specified below:
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -303,6 +303,15 @@ CPU-ATTR    := 'cpu=' <valid-cpu-name>
 TUNE-ATTR   := 'tune=' <valid-tune-name>
 ```
 
+The target attribute does not support multi-versioning. The compiler should
+emit an error if a function is defined more than once. For example, the
+following code should trigger an error because foo is declared twice:
+
+```
+__attribute__((target("arch=+v"))) int foo(void) { return 0; }
+__attribute__((target("arch=+zbb"))) int foo(void) { return 1; }
+```
+
 ## Intrinsic Functions
 
 Intrinsic functions (or intrinsics or built-ins) are expanded into instruction sequences by compilers.
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -248,7 +248,6 @@ FULLARCHSTR            := <full-arch-string>
 EXTENSION              := <OP> <EXTENSION-NAME> <VERSION>
 
 OP                     := '+'
-                        | '-'
 
 VERSION                := [0-9]+ 'p' [0-9]+
                         | [1-9][0-9]*
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -254,8 +254,6 @@ takes effect. For example, `arch=+zbb;arch=+zba` will be equivalent to
 The interactions between the attribute and the command-line option are
 specified below:
 
-Here's a refined version of your statement:
-
 - `arch=`: Its behavior depends on the syntax used:
            1) Adding extra extensions: It will merge the extension list with the `-march` option.
            2) If a full architecture string is specified by `arch=`, it will override the `-march` option.
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -208,8 +208,20 @@ This attribute is incompatible with the `naked` attribute.
 `target` attribute used for enable or disable a set of features or extensions
 for the fucntion.
 
+For instance, you can enable the `v` extension for a specific function even if
+`-march` or `-mcpu` does not include the `v` extension. Additionally,
+this won't alter the global setting.
+```c
+__attribute__((target("+v")))
+int
+foo (int a)
+{
+  return a + 5;
+}
+```
+
 Function with target attribute should not affect the file scope build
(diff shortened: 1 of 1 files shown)


kito-cheng:
cc @asb @luismarques @maskray @jrtc27 @topperc @preames @palmer-dabbelt  @vineetgarc

kito-cheng:
PoC for LLVM:  https://reviews.llvm.org/D151730

palmer-dabbelt:
Is this meant to plumb into function multi versioning as well, or just control single-versioned functions?  Either way Jeff might have something to say, not sure what his github ID is though...

kito-cheng:
> Either way Jeff might have something to say, not sure what his github ID is though...

Should be this one @JeffreyALaw

kito-cheng:
Changes:

- Drop the '-' op due to lack of clear definition.

kito-cheng:
Changes:

 - Fix the syntax to allow more than two ATTR.
 - Speicfy the interactions among `arch`, `tune`, and `cpu`.
 - Describe the behavior if same ATTR appear more than once.
 - Mention compiler should always generate the function even the
      attribute compiler will generate same code without attribute.

TODO:
- We may specify the inlining rule like other target
https://gcc.gnu.org/onlinedocs/gcc/AArch64-Function-Attributes.html#Inlining-rules

riscv-c-api.md
```diff
@@ -170,6 +170,90 @@ For example:
...
+### `__attribute__((target("<ATTR-STRING>")))`
+
+`target` attribute used for enable or disable a set of features or extensions
+for the fucntion.
+
+For instance, you can enable the `v` extension for a specific function even if
+`-march` or `-mcpu` does not include the `v` extension. Additionally,
+this won't alter the global setting.
+```c
+__attribute__((target("+v")))
```
luismarques, line 182:
```suggestion
__attribute__((target("arch=+v")))
```
Is this what was intended?

riscv-c-api.md
```diff
@@ -170,6 +170,90 @@ For example:
...
+int
+foo (int a)
+{
+  return a + 5;
+}
+```
+
+Function with target attribute should not affect the file scope build
+attributes, i.g. File complied with `-march=rv64ima` and a function declares with
```
luismarques, line 191:
```suggestion
attributes, e.g. File compiled with `-march=rv64ima` and a function declares with
```

riscv-c-api.md
kito-cheng: Oh, you catch a bug, it's not intentionally

riscv-c-api.md
```diff
@@ -170,6 +170,90 @@ For example:
...
+`-mtune`, and `-mcpu` options. The `cpu` can be treated as a combination of
+`arch` + `tune` but has a lower priority than the other two. For instance,
+`cpu=sifive-u74` is equivalent to `arch=rv64gc` and `tune=sifive-7-series`.
+However, if `arch=` or `tune=` values are provided, they will override the `cpu`
+value. Therefore, `cpu=sifive-u74;arch=rv64g` will be equivalent to
+`arch=rv64g;tune=sifive-7-series`, and `cpu=sifive-u74;tune=sifive-5-series`
+will be equivalent to `arch=rv64gc;tune=sifive-5-series`.
+
+If the same type of attribute is given more than once, only the latest one
+takes effect, e.g. `arch=+zbb;arch=+zba` will be equivalent to `arch=+zba`.
```
preames, line 217: What is the effect of having -march at the command line and a march on the function which is a subset of the command line?  I think either disallowing or full replacement is fine, we just don't want to try merging.

If I specific march at the command line, and mcpu via the attribute, what happens?  mcpu would normally be lower priority, but is this still true comparing command line and attribute?

preames (changes requested):

riscv-c-api.md
```diff
@@ -170,6 +170,90 @@ For example:
...
+}
+```
+
+Function with target attribute should not affect the file scope build
+attributes, e.g. File compiled with `-march=rv64ima` and a function declares with
+`__attribute__((target("arch=+zbb")))`, the `Tag_RISCV_arch` build attribute
+should be `rv64ima` rather than `rv64ima_zbb`.
+
+The compiler may emit mapping symbol at the beginning of the function with
+target attribute if the function use different set of ISA extension.
```
topperc, line 196: use -> uses

riscv-c-api.md
```diff
@@ -170,6 +170,90 @@ For example:
...
+
+`<ATTR-STRING>` can be specify following target attributes:
```
topperc, line 198: "can be specify following" -> "can specify the following"

riscv-c-api.md
```diff
@@ -170,6 +170,90 @@ For example:
...
+                        |
+
+EXTENSION-NAME         := Naming rule is defined in RISC-V ISA manual
+
+CPU-ATTR    := 'cpu=' <valid-cpu-name>
+TUNE-ATTR   := 'tune=' <valid-tune-name>
+```
+
+NOTE: The `target` attribute won't effect function name mangling for both C and C++.
+NOTE: The compiler should always generate the function with the `target`
```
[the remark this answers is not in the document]
kito-cheng: Add one example, @BeMg told me some case will cause undefined reference due to the attribute have no effect.

riscv-c-api.md
kito-cheng: Add one more paragraph to describe the interaction between attribute and option.

`arch=` has two different behavior here: full-arch-str will override `-march`, but `+ext` will combine with `-march`.
`cpu=` will override `-march` if `arch=` not given.

riscv-c-api.md
```diff
@@ -170,6 +170,90 @@ For example:
...
+                        | [1-9][0-9]*
+                        |
+
+EXTENSION-NAME         := Naming rule is defined in RISC-V ISA manual
+
+CPU-ATTR    := 'cpu=' <valid-cpu-name>
+TUNE-ATTR   := 'tune=' <valid-tune-name>
+```
+
+NOTE: The `target` attribute won't effect function name mangling for both C and C++.
```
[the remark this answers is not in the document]
kito-cheng: dropped

kito-cheng:
Changes:
- Clarify the interaction between attribute and option, e.g. `-march` and `arch=`.
- Improving wording and fix several  grammatical issue

kito-cheng:
Minor tweaks

- Drop `disable a set of features or extensions`, we don't have syntax
  to drop now.
- Tweak description for `arch=`: describe two different cases of `arch=`
  clearly.
- Drop last sentence, that will happened on multi-version funciton but
  not happen in `target` attribute.

riscv-c-api.md
kito-cheng: Drop for now

kito-cheng:
cc @maskray, who might interested on this topic

riscv-c-api.md
```diff
@@ -203,6 +203,102 @@ function declares more than one mode or an undefined mode.
...
+`arch=rv64gc;tune=sifive-5-series`.
+
+If the same type of attribute is specified more than once, only the latest one
+takes effect. For example, `arch=+zbb;arch=+zba` will be equivalent to
+`arch=+zba`.
+
+The interactions between the attribute and the command-line option are
+specified below:
+
+Here's a refined version of your statement:
```
cmuellner, line 257: This line looks like it should not be here.

cmuellner (approved):
LGTM.

We could reach out to sig-toolchains an announce that this will be merged in a few days, so that people can provide feedback if they are interested.

cmuellner:
I've reached out to the sig-toolchains list with a request for review with the intend to merge this PR in two weeks.

The planned timeline is as follows:

Oct 13: [Send request for review to sig-toolchains list](https://lists.riscv.org/g/sig-toolchains/message/651)
Oct 27: End of review period (PR will be merged if there are no outstanding issues)

riscv-c-api.md
kito-cheng: Yeah...thanks catching that, removed

preames:
To provide a bit more context on the in person conversation last week...

@kito-cheng and I sat down at the LLVM developers meeting to go over the then current draft text and identified a couple of last minor issues.  I provided a verbal LGTM at that point, subject to the changes identified being made.  @topperc and @maskray joined in for the last part of our spoken conversation just before I ran off to a talk.  I'm unclear if all their concerns were resolved.  I think they were, but I don't want to speak for them.  

One point I want to call out in our discussion is that we tentatively decided *not* to implement function multi-version semantics for target in C++.  (On X86, C and C++ semantics for this attribute differ with C++ providing a form of a overloading based on attribute.)  We're going to leave that case to target_clone and keep the plain target attribute simpler.  This does result in a difference with other architectures, but given aarch64 and x64 already seem to differ significantly on the semantics assigned to this attribute, that seemed acceptable.

riscv-c-api.md
```diff
@@ -203,6 +203,100 @@ function declares more than one mode or an undefined mode.
...
+
+```c
+__attribute__((target("arch=+v")))
+int foo(int a)
+{
+  return a + 5;
+}
+```
+
+Using the `target` attribute for a function should not affect the file scope
```
topperc, line 223: "file scope" should maybe be "translation unit scope" since files can include other files. The global attributes apply to every file being included.

riscv-c-api.md
```diff
@@ -203,6 +203,100 @@ function declares more than one mode or an undefined mode.
...
+build attributes. For example, if a file is compiled with `-march=rv64ima` and
+a function is declared with `__attribute__((target("arch=+zbb")))`, the
+`Tag_RISCV_arch` build attribute should remain `rv64ima`, not `rv64ima_zbb`.
+
+The compiler may emit a mapping symbol at the beginning of a function with the
```
topperc, line 228: Is there some documentation we can link "mapping symbol" to here?

riscv-c-api.md
```diff
@@ -203,6 +203,100 @@ function declares more than one mode or an undefined mode.
...
+the `-march`, `-mtune`, and `-mcpu` options. The `cpu` attribute can be seen as
+a combination of `arch` + `tune` but holds a lower priority than the other two.
+For instance, `cpu=sifive-u74` equates to `arch=rv64gc` and
+`tune=sifive-7-series`. However, if values for `arch=` or `tune=` are provided,
+they will override the `cpu` value. Therefore, `cpu=sifive-u74;arch=rv64g` is
+equivalent to `arch=rv64g;tune=sifive-7-series`, and
+`cpu=sifive-u74;tune=sifive-5-series` is equivalent to
+`arch=rv64gc;tune=sifive-5-series`.
+
+If the same type of attribute is specified more than once, only the latest one
```
topperc, line 250: Is this useful behavior or should we just reject if the same type appears twice? Do we envision building a string through preprocessor macros where it might not be obvious to the programmer that the same value appears twice?

kito-cheng:
Reach consensus on error when multiple `__attribute__((target))` and duplicated value (same type of attribute) at RISC-V LLVM sync meeting.

Action item: Kito will discuss in next RISC-V GCC sync meeting, make sure everyone is happy with that.

MaskRay:
For x86, GCC in C++ mode (but not in C mode) enables function multi-versioning for the following setup:
```
__attribute__((target("default"))) int foo(void) { return 0; }
__attribute__((target("sse4.2"))) int foo(void) { return 1; }
```

I think this is confusing. For RISC-V I think it makes sense not to enable function multi-versioning or make `__attribute__((target("sse4.2"))) ` part of the mangling.

kito-cheng:
Changes:
- Compiler emit error instead of take latest one if same type of attribute specified more thane once, also same behavior for target attribute specified twice.
- Explicitly declare `target` attribute not support multi-versioning.

preames:
This has been under active discussion for months, and recently went through the 2 week comment period without significant comment.  I think it's time to move forward here.

kito-cheng:
Let moving forward, thanks everyone :)

PR #38, opened 2023-04-27
Add inline asm operand constraints for vector register

kito-cheng:
We have 3 common operand constraints constraint between GCC and LLVM here:

- vr: Any vector register, v0-v31
- vd: Any vector register, excluding v0, used for avoid overlapping with mask register.
- vm: Mask vector register, only v0.


cmuellner:
The description of this PR mixes 'vd' and 'vm'.
However, the actual change in the commit is correct and corresponds to the GCC implementation.
LGTM.

PR #39, opened 2023-04-27
Introduce new constraint for register pair

### LINKED ISSUE #37: Asm constraint to require an even register
We don't currently have an asm constraint to require an even register (for an input or output in a register pair) and I don't believe there's a cross-target constraint already defined. We should likely decide on one. This could be used with the proposed zacas extension (amocas.d on rv32 or amocas.q on rv64), for zdinx, and IIRC the P extension.

kito-cheng:
Must start with an even-numbered register.

NOTE: This is *NOT* implemented on either on LLVM or GCC yet, this PR will hold until at least one compiler has implemented.

Fix #37

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -225,6 +225,7 @@ statements, including both RISC-V specific and common operand constraints.
 | m                  | An address that is held in a general-purpose register with offset.      |      |
 | A                  | An address that is held in a general-purpose register.      |      |
 | r                  | General purpose register                      |      |
+| R                  | Register pair for GPR, and it must start with an even-numbered register.   |      |
 | f                  | Floating-point register                       |      |
 | i                  | Immediate integer operand                     |      |
 | I                  | 12-bit signed immediate integer operand      |      |


cmuellner:
There is a decision between two options:
* an even-numbered register
* a register pair that is identified by the even, lower-numbered register

The difference is, that the second implies the amount of registers in the constraint, while the other one does not (or implies 1 if not stated otherwise).

#37 requests the first option. This PR introduces the second. Is there a reason to go for the second option besides personal preference (I admit that I prefer the second option for this reason)?

And would clobbering a register with the constraint "G" imply that a register pair is clobbered (is this easy to implement in GCC/LLVM)?

lenary:
I see my proposal in #92 conflicts with this, as it proposes a different name (`Pr` vs `R`). I chose `Pr` because I felt it was extensible to other kinds of paired registers in the future, potentially.

kito-cheng:
@lenary Yeah, I agree this should be designed for more extensibility, and your proposal is better than this one :) so let me drop this proposal to prevent misleading that we have two competing proposals.

Issue #32, opened 2023-02-07
[RFC] Preprocessor definition to specify data alignment strict(ness)


vineetgarc:
I would like to propose a compiler generated preprocessor macro to signal data alignment heuristics used by the compiler for codegen.
This could be used by library writers to target code for efficient unaligned access (or not).
Granted we would like to eventually use runtime probing to figure this out, it would help the user to know what the compiler assumptions are anyways (or if user coaxed it into a certain semantics via additional toggles)

At a high level it reflects whether gcc toggle -mstrict-align has been used to build.

Speaking of gcc there' another wrinkle to worry about. gcc driver has a notion of cpu tune param which specifies whether unaligned accesses are generally efficient on the cpu (or not). And a cpu tune with slow_unaligned_access=true will disregard -mno-strict-align,   compiler codegen assuming as if -mstrict-align was passed. So the proposed preprocessor macro needs to also reflect this.

Here's the specification:
**__riscv_strict_align**  (only defined if -mstrict-align or cpu tune param slow_unaligned_access=true)
Value 1: if -mstrict-align
value 2: if cpu tune param specifies slow_unaligned_access = true

vineetgarc:
FWIW there's a gcc patch posted along those lines
https://gcc.gnu.org/pipermail/gcc-patches/2023-January/610115.html

asb:
Thanks for opening this discussion. One thing I'm wondering about is whether it makes sense to couple this primarily to -mstrict-align / -mno-strict-align.

The RISC-V profiles proposal finally introduces a way to indicate whether misaligned loads/stores are supported or not (Zicclsm, which once supported by compilers should be queryable via `__riscv_zicclsm` as for any other extension).

Once you have that capability, might it be more intuitive to have a define dedicated solely to whether misaligned loads/stores are slow or fast? Not defined if Zicclsm isn't present, 1 if slow (potentially also if -mstrict-align was used, taking that as a hint that such accesses should be considered slow), 0 if not slow.

I'm wary of derailing this by trying to over-generalise it. But it's probably also worth at least considering if we'd like some naming scheme that reflected the extension that tuning hint is related to. e.g. `__riscv_zicclsm_slow`.

vineetgarc:
> Thanks for opening this discussion. One thing I'm wondering about is whether it makes sense to couple this primarily to -mstrict-align / -mno-strict-align.

[...]

PR #40, opened 2023-05-10
Add macros for unaligned access

vineetgarc:


nick-knight (commented):
The RISC-V ISA spec (and profiles, etc.) appear to consistently use the word "misaligned". Is there a reason this proposal uses "unaligned"?

vineetgarc:
> The RISC-V ISA spec (and profiles, etc.) appear to consistently use the word "misaligned". Is there a reason this proposal uses "unaligned"?

No real reason, just the terminology I'm more used to from past. I can certainly respin with new verbiage.

riscv-c-api.md
```diff
@@ -94,6 +97,15 @@ The value of `__riscv_v_elen_fp` is defined by the following rules:
 If multiple rules apply, the maximum value is taken.
 If none of the rules apply, `__riscv_v_elen_fp` is undefined.
 
+### __riscv_misaligned_{fast,slow,bad}
+
+These can be used in common library code to compile time segregate code which relies
+on misaligned access being fast or not.
+A typical complier could (but not necessarily) map fast variant to -mno-strict-align
+and bad to -mstrict-align, if specified.
```
kito-cheng, line 105: I would like to drop that since clang has different behavior here, it just accept and ignore `-mno-strict-align` and `-mstrict-align` for RISC-V clang.

Or move it become a note to say GCC is using that way to decide which macro should be defined.

kito-cheng:
cc @asb @luismarques

riscv-c-api.md
vineetgarc: Sorry not clear what you mean: clang ignores either of them ? and uses a different toggle to affect misaligned access codegen or doesn't support that at all (extremely unlikely)

kito-cheng: > Sorry not clear what you mean: clang ignores either of them ? and uses a different toggle to affect misaligned access codegen or doesn't support that at all (extremely unlikely)

Yes, RISC-V clang accept that and then ignore that for now - clang did support misaligned code gen and some more are coming up[1], option handling in clang part can easily be implement/fix later.

Anyway just second thought about this: This should not major blocker issue here, just a minor implementation issue, and you already mention `(but not necessarily)`.

kito-cheng (commented):
LGTM as a RISC-V GCC maintainer, but I would like to wait LLVM folks ack here.

vineetgarc:
> I think from #32, @aswaterman was concerned that `_bad` was unnecessarily perjorative and `_avoid` might be preferable.

Fair enough. Updated pull request with bad -> avoid
 
> I had a slight preference for zicclsm even if it's not very comprehensible just to avoid introducing too many ways of saying the same thing.

I tend to disagree still, but we can agree to disagree :-)

> But I don't feel strongly enough to block this. I think we should wait for @aswaterman's feedback though, as he participated in the other thread.

Sure thing.

kito-cheng (approved):
LGTM, and I am OK with those name

ping  @aswaterman

aswaterman:
I'm OK with this scheme.  I guess I agree with @asb in principle, but, like him, I don't think it's actually a problem, so I don't want to hold this up.

asb:
Thanks Andrew. Sounds like we have consensus on this then? LGTM.

kito-cheng:
Thanks @vineetgarc , gonna merge this :)

PR #42, opened 2023-06-13
Save CSRs in interrupt functions

palmer-dabbelt:
Link: https://inbox.sourceware.org/gcc-patches/20230613064126.1323-1-jinma@linux.alibaba.com/


kito-cheng (approved):
LGTM, I guess I missed this before, and I think it's totally reasonable.

nick-knight:
I think the original intent was to save/restore all (ISR-modified) architectural state associated with these extensions, which in these cases means the data registers and CSRs. I guess the confusion is that the doc just said "floating point/vector register[s]", which could be misread as only referring to the data registers, and excluding the CSRs. I think this clarification could be implemented by adding a few words to the previous sentence, rather than adding another one. But it's fine.

PR #44, opened 2023-06-30
Scalar bitmanip and crypto intrinsics proposal

topperc:
This is a proposal for scalar crypto intrinsics for the Zk* extensions.

I've removed the recommendation to use 'long' for XLEN. There's nothing guaranteeing that 'long' is XLEN. For example, if we were to support ILP32 on RV64, 'long' would become 32 bits while XLEN would still be 64 bits.

Other design decisions are spelled out in the document.

Some text has been copied from #23.

One open question is whether we should emulate 64-bit intrinsics on RV32 where it is feasible.  This could be useful for the sha512 intrinsics and some others.

@cmuellner @kito-cheng @mjosaarinen

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -391,18 +391,6 @@ Sign extension of 32-bit values on RV64 is not reflected in the interface.
 | `uint32_t __riscv_sm4ed(uint32_t rs1, uint32_t rs2, const int bs);`     | `sm4ed`       | Zksed             | `bs`=[0..3] |
 | `uint32_t __riscv_sm4ks(uint32_t rs1, uint32_t rs2, const int bs);`     | `sm4ks`       | Zksed             | `bs`=[0..3] |
 
-### Cryptography Intrinsics Implementation Guarantees
-
-The `riscv_crypto.h` can implement the intrinsics in many ways
-(early implementations used inline assembler). Builtin mapping is a
-compiler and system specific issue.
-
-Due to the data-independent latency ("constant time") assertions of
-the `Zkt` extension, the header file or the compiler can't use table
-lookups, conditional branching, etc., when implementing crypto intrinsics.
-In production (cryptographic implementations), the execution latency of
-all cryptography intrinsics must be independent of input values.
-
 ## Constraints on Operands of Inline Assembly Statements
 
 This section lists operand constraints that can be used with inline assembly
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -288,18 +288,17 @@ enum {
 | `__RISCV_NTLH_INNERMOST_SHARED`  | `ntl.s1`    |
 | `__RISCV_NTLH_ALL`               | `ntl.all`   |
 
-### Scalar Cryptography Extension Intrinsics
+### Scalar Bit Manipulation Extension Intrinsics
 
-In order to access the RISC-V scalar crypto intrinsics, it is necessary to
-include the header file `riscv_crypto.h`.
+In order to access the RISC-V scalar bit manipulation intrinsics, it is
+necessary to include the header file `riscv_bitmanip.h`.
 
 The functions are only only available if the compiler's `-march` string
 enables the required ISA extension. (Calling functions for not enabled
 ISA extensions will lead to compile-time and/or link-time errors.)
 
 Intrinsics operating on XLEN sized value are not available as there is no type
-defined. If `xlen_t` is added in the future, this can be revisited. It is
-assumed that cryptographic code will be written using fixed bit widths.
+defined. If `xlen_t` is added in the future, this can be revisited.
 
 Unsigned types are used as that is the most logical representation for a
 collection of bits.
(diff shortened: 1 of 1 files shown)


riscv-c-api.md
```diff
@@ -227,14 +226,99 @@ TYPE ::= Optional type postfix.
...
+| `uint64_t __riscv_brev8_64(uint64_t x);`                                | `brev8`              | Zbkb (RV64)       | |
+| `uint32_t __riscv_zip_32(uint32_t x);`                                  | `zip`                | Zbkb (RV32)       | No emulation for RV64 |
+| `uint32_t __riscv_unzip_32(uint32_t x);`                                | `unzip`              | Zbkb (RV32)       | No emulation for RV64 |
+| `uint32_t __riscv_clmul_32(uint32_t x);`                                | `clmul`              | Zbc, Zbkc         | Emulated with `clmul`+`sext.w` on RV64 |
+| `uint64_t __riscv_clmul_64(uint64_t x);`                                | `clmul`              | Zbc, Zbkc (RV64)  |                                  |
+| `uint32_t __riscv_clmulh_32(uint32_t x);`                               | `clmulh`             | Zbc, Zbkc         | Emulated with `clmul`+`srai` on RV64 |
+| `uint32_t __riscv_clmulh_32(uint32_t x);`                               | `clmulh`             | Zbc, Zbkc         | Emulated with `clmul`+`srai` on RV64 |
```
kito-cheng, line 273: Duplicated entry or it should be `__riscv_clmulh_64`?

topperc: Fixed. Thanks.

topperc:
@nick-knight @Xinlong-Wu @aswaterman @ChunyuLiao

Xinlong-Wu:
here is a related issue: https://github.com/rvkrypto/rvkrypto-fips/issues/11

kito-cheng (approved):
I support those intrinsic functions explicitly use the fixed length integer instead of xlen, that could prevent user to use wrong length to implement stuffs :)

ChunyuLiao:
LGTM, thanks.

topperc:
I removed `__riscv_clmulh_32` and `__riscv_clmulr_32` from RV64 as the emulation is 4-6 instructions.

topperc:
I fixed the number of arguments to the xperm intrinsics. And I reduce the rotate amount for `__riscv_rol_64` and `__riscv_ror_64` intrinsics to `uint32_t`

riscv-c-api.md
```diff
@@ -227,14 +226,125 @@ TYPE ::= Optional type postfix.
...
+
+### Cryptography Intrinsics Implementation Guarantees
+
+The `riscv_crypto.h` can implement the intrinsics in many ways
+(early implementations used inline assembler). Builtin mapping is a
+compiler and system specific issue.
+
+Due to the data-independent latency ("constant time") assertions of
+the `Zkt` extension, the header file or the compiler can't use table
+lookups, conditional branching, etc., when implementing crypto intrinsics.
```
jrtc27, line 344: Is this actually a useful guarantee? The compiler can do whatever it likes around the intrinsic. If you need data-independent timing then you need to write the entire thing in assembly...

wangpc-pp:
What about adding a intrinsic that reads `seed` CSR defined in `Zkr`?

topperc:
> What about adding an intrinsic that reads `seed` CSR defined in `Zkr`?

It's not guaranteed to be accessible to user level software. I don't think it belongs in this proposal.

topperc:
What do we need to do to move this proposal forward?

pz9115 (approved):
LGTM, I think we have some bitmanip intrinsic work to do in GCC part, thank you.

nick-knight (approved):

cmuellner (approved):
LGTM

Thank you very much work working this out!

cmuellner:
We have discussed this in today's SIG toolchain call.
We have positive feedback in the comments here and the PR had much time so everyone had much time to give comments.
Therefore, I've reached out to the sig-toolchains list with a request for review with the intend to merge this PR in two weeks 

The planned timeline is as follows:
* Oct 9: [Send request for review to sig-toolchains list](https://lists.riscv.org/g/sig-toolchains/message/648)
* Oct 23: End of review period
* October: Work on patches for the intrinsics (GCC, LLVM)
* End of October: merge PR if patches are on the list (GCC) or in a PR (LLVM)

topperc:
Clang/llvm patch https://reviews.llvm.org/D155647

nick-knight:
Are we adding new macros like `__riscv_zbkb`? If so, is it important to add them to this document?

cmuellner:
> Are we adding new macros like `__riscv_zbkb`? If so, is it important to add them to this document?

We have a list in this repo, but it does not make sense to add each and every extension manually (which is the reason that nobody does it).
Kito attempted to address this in https://github.com/riscv-non-isa/riscv-c-api-doc/pull/28, but the PR was not finished.

So instead of a PR that adds `__riscv_zbkb` it would be great to get a more generic description of the test macros.

cmuellner:
@pz9115 is working on the GCC patches.
The work is tracked in an dev-partners issue: https://github.com/riscv-admin/dev-partners/issues/37

topperc:
I think the public review period for this is over now. How do we proceed?

cmuellner:
@pz9115 is there an ETA for the GCC patch?

pz9115:
Still WIP, plan to send patches within this month.

kito-cheng:
@cmuellner I would suggest moving forward this PR, I mean merge this first, I believe the proposal isn't come with complicated technical issue on the GCC side, just need few more time to implement.

Also I am happy to accept that on GCC 14 even it's not stage 1.

cmuellner:
I kept it open, because I was not sure if this is still being work on.
Now, that this has been confirmed in yesterday's SIG Toolchain call, it is probably the best to merge it.

Thanks @topperc for the PR and the patience!

PR #46, opened 2023-07-13
Locality for prefetch built-in proposal

BeMg:
ARM[1] provide the its own semantic on prefetch built-in locality.

It seem we could do the same thing in RISC-V.

[1] https://developer.arm.com/documentation/101458/2010/Coding-best-practice/Prefetching-with---builtin-prefetch

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -402,15 +402,15 @@ enum {
 
 ### Prefetch Intrinsics
 
-The zicbop extension provide the prefetch instruction to allow users to optimize data access patterns by providing hints to the hardware regarding future data accesses. It is supported through a compiler-defined built-in function with three arguments that specify its behavior.
+The Zicbop extension provide the prefetch instruction to allow users to optimize data access patterns by providing hints to the hardware regarding future data accesses. It is supported through a compiler-defined built-in function with three arguments that specify its behavior.
 
 ```
 void __builtin_prefetch(const void *addr, int rw, int locality)
 ```
 
-The locality for the built-in __builtin_prefetch function in RISC-V can be achieved using the Non-Temporal Locality Hints (NTLH) extension. According to Non-Temporal Locality Hints extension, it indicates that a cache line should be prefetched into a cache that is outer from the level specified by the NTL when a NTL instruction is applied to prefetch instruction.
+The locality for the built-in `__builtin_prefetch` function in RISC-V can be achieved using the Non-Temporal Locality Hints (Zihintntl) extension. According to Non-Temporal Locality Hints extension, it indicates that a cache line should be prefetched into a cache that is outer from the level specified by the NTL when a NTL instruction is applied to prefetch instruction.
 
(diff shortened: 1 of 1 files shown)


kito-cheng:
cc @asb @topperc @palmer-dabbelt @JeffreyALaw

kito-cheng:
> This also exposes an issue with our treatment of the 'zihintntl' extension on the LLVM side at least (I'm not sure if the same issue is there in GCC). The proposed prefetch locality patch for LLVM only emits the ntl hints if the zihintntl extension is known to be present. We should probably emit the add x0, x0, x? based mnemonics directly even if zihintntl isn't known to be present.

GCC implementation also require `zihintntl` too, that also remind another issue I discussed with @topperc here is: should we allow those hint mnemonics still available even corresponding is not enabled? And actually prefetch is also a hint instruction too, so we might did the same treatment for `Zicbop` too.

 I guess this worth to put to the next meeting agenda again.

riscv-c-api.md
```diff
@@ -235,6 +235,25 @@ long __riscv_clmul (long a, long b); // clmul rd, rs1, rs2
 vint8m1_t __riscv_vadd_vv_i8m1(vint8m1_t vs2, vint8m1_t vs1, size_t vl); // vadd.vv vd, vs2, vs1
 ```
 
+### Prefetch Intrinsics
+
+The Zicbop extension provide the prefetch instruction to allow users to optimize data access patterns by providing hints to the hardware regarding future data accesses. It is supported through a compiler-defined built-in function with three arguments that specify its behavior.
```
hiraditya, line 240: nit s/provide/provides

riscv-c-api.md
```diff
@@ -235,6 +235,25 @@ long __riscv_clmul (long a, long b); // clmul rd, rs1, rs2
...
+
+```
+void __builtin_prefetch(const void *addr, int rw, int locality)
+```
+
+The locality for the built-in `__builtin_prefetch` function in RISC-V can be achieved using the Non-Temporal Locality Hints (Zihintntl) extension. According to Non-Temporal Locality Hints extension, it indicates that a cache line should be prefetched into a cache that is outer from the level specified by the NTL when a NTL instruction is applied to prefetch instruction.
```
hiraditya, line 246: nit:
> According to Non-Temporal Locality Hints extension, it indicates that a cache line should be ...
The Non-Temporal Locality (NTL) Hints extension indicates that a cache line should be ...

hiraditya:
Thanks for addressing the changes i requested.

kito-cheng (approved):
LGTM, @asb do you mind take a look again? :)

cmuellner (approved):

BeMg:
Resolve merge conflict.

kito-cheng:
Although @asb is not response yet, but he has aware this PR and patch for LLVM[1], also it merged into LLVM trunk, so I treat it as no objection from LLVM community and moving this PR forward.

[1] https://reviews.llvm.org/D154691

Issue #30, opened 2022-07-27
Intrinsics support for Zihintntl extension


aswaterman:
The Zihintntl extension has recently passed AR.  The spec is here: https://github.com/riscv/riscv-isa-manual/blob/10eea63205f371ed649355f4cf7a80716335958f/src/zihintntl.tex

During the AR, we wanted to raise the issue of whether and how the extension would be exposed in the RISC-V C API.  Can y'all ponder the following and opine?

In x86, for example, `_mm_stream_pi` (https://github.com/gcc-mirror/gcc/blob/e75da2ace6b6f634237259ef62cfb2d3d34adb10/gcc/config/i386/xmmintrin.h#L1279-L1291) is roughly equivalent to `c.ntl.all; sd` in RISC-V.

(ARMv8 has LDNP/STNP instructions, but I couldn't find an intrinsics mapping for them.)

Zihintntl is more general than x86's solution in a few dimensions:
- NTL hints can be used on both loads and stores (in x86, there are only non-temporal stores, AFAIK)
- NTL hints can be used on any kind of load and store (in x86, it's only MMX/SSE/AVX, AFAIK)
- NTL hints can express a memory hierarchy level (x86 can only express the equivalent of `ntl.all`, AFAIK)

With this in mind, the questions for the RISC-V C API folks are: how do we expose this facility in the RISC-V C API?  How much of its generality do we expose?  Do you foresee any impediments?

cc @kito-cheng @ptomsich

cmuellner:
The specification requires the memory access to be the "immediately subsequent instruction".
This is hard/impossible to guarantee by an intrinsic that does not include the memory access operation.
Therefore I would include the memory access in the API.

I see two possible solutions.

Proposal one: provide NTL loads and stores (similar to the atomics builtins):
```
 type __riscv_ntl_load (type *ptr, enum ntl_domain domain);
void __riscv_ntl_store (type *ptr, enum ntl_domain domain);
```
However, this will probably only work reasonably well for single-GPR/FPR-memory transfers. E.g. vector memory accesses probably need a _ntl variant of the vector load/store intrinsics.

Proposal two: no intrinsics, but a NTL function attribute that [does not block inlining](https://gcc.gnu.org/bugzilla//show_bug.cgi?id=101279)
```
// all memory accesses in this function emit an NTL hint before the memory access instruction
__attribute__((target("ntl_domain=DOMAIN"))
static inline void
[...]

PR #47, opened 2023-07-19
zihintntl Intrinsic function proposal

BeMg:
This is a proposal for zihintntl intrinsic function.

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -242,7 +242,7 @@ The RISC-V zihintntl extension provides the RISC-V specific intrinsic functions 
 In order to access the RISC-V NTLH intrinsics, it is necessary to
 include the header file `riscv_ntlh.h`.
 
-The functions are only only available if the compiler enables the zihintntl extension.
+The functions are only available if the compiler enables the zihintntl extension.
 
 ```
 type __riscv_ntl_load (type *ptr, int domain);
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -249,6 +249,13 @@ type __riscv_ntl_load (type *ptr, int domain);
 void __riscv_ntl_store (type *ptr, type val, int domain);
 ```
 
+There are overloaded functions of `__riscv_ntl_load` and `__riscv_ntl_store`. When these intrinsic functions omit the `domain` argument, the `domain` is implied as `__RISCV_NTLH_ALL`.
+
+```
+type __riscv_ntl_load (type *ptr);
+void __riscv_ntl_store (type *ptr, type val);
+```
+
 The types currently supported are:
 
 - Integer types.
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -235,6 +235,44 @@ long __riscv_clmul (long a, long b); // clmul rd, rs1, rs2
 vint8m1_t __riscv_vadd_vv_i8m1(vint8m1_t vs2, vint8m1_t vs1, size_t vl); // vadd.vv vd, vs2, vs1
 ```
 
+### NTLH Intrisics 
+
+The RISC-V zihintntl extension provides the RISC-V specific intrinsic functions for generating non-temporal memory accesses. These intrinsic functions provide the domain parameter to specify the behavior of memory accesses.
+
+In order to access the RISC-V NTLH intrinsics, it is necessary to
+include the header file `riscv_ntlh.h`.
+
+The functions are only only available if the compiler enables the zihintntl extension.
+
+```
+type __riscv_ntl_load (type *ptr, int domain);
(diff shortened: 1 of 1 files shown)


asb:
For further context, I understand this intends to match what Clang currently provides (which is gated on enabling experimental extension support right now).

kito-cheng:
For more context: this also match what @cmuellner proposed https://github.com/riscv-non-isa/riscv-c-api-doc/issues/30#issuecomment-1200199130 before

kito-cheng:
c.c. @aswaterman

aswaterman:
LGTM.  The only comment I'd make is that we'd consider adding overloaded versions that omit `domain`, implying `ALL`.  But I don't know offhand whether overloaded intrinsics are valid in C, as opposed to C++, and maybe this "easy-mode" will prove controversial for other reasons--so if there's disagreement, I don't want to block forward progress.

nick-knight:
> But I don't know offhand whether overloaded intrinsics are valid in C (as opposed to C++),

Both GCC's and Clang's C-language dialects support a limited form of C++ function overloading. We use this already in the vector intrinsics ("overloaded API"). Since these are the only compilers we require to prove concepts, this seems actionable here as well.

BeMg:
> LGTM. The only comment I'd make is that we might consider adding overloaded versions that omit the `domain` argument, implying `ALL`. But I don't know offhand whether overloaded intrinsics are valid in C (as opposed to C++), and maybe this "easy-mode" will prove controversial for other reasons--so if there's disagreement, I don't want to block forward progress.

The version that omit the `domain` argument I think the clang exists the similar built-in function `__builtin_nontemporal_load/store` [1]. 

[1] https://clang.llvm.org/docs/LanguageExtensions.html#non-temporal-load-store-builtins

aswaterman:
@BeMg OK.  Can we document that API here, including that it implies `ALL`, so that GCC and LLVM will match?  Or is the problem that we can't make it part of this API because it does not include the `__riscv` prefix?

BeMg:
I'm not sure whether to add the `__builtin_nontemporal_load/store` to proposal. Besides it doesn't not including the `__riscv` prefix, GCC seem doesn't has these built-in functions now.

cc @kito-cheng 

The overloaded version look like more consistent, so I add it into proposal. 

It could be implemented by some macros. 

```
#define __riscv_ntl_load_with_domain(PTR, DOMAIN) __builtin_riscv_ntl_load((PTR), (DOMAIN))
#define __riscv_ntl_load_without_domain(PTR) __builtin_riscv_ntl_load((PTR), __RISCV_NTLH_ALL)

#define SELECT_NTL_LOAD(_1,_2,NAME,...) NAME
#define __riscv_ntl_load(...) SELECT_NTL_LOAD(__VA_ARGS__, __riscv_ntl_load_with_domain, __riscv_ntl_load_without_domain)(__VA_ARGS__)
```

kito-cheng (approved):
LGTM, seems like the overloading trick is work both on GCC and clang/LLVM, and I will try to find some resource to implement GCC part.

I would like to get explicitly ack from @aswaterman and @asb before merge.

aswaterman:
OK with me, too.

BeMg:
I will implement the overloaded version in LLVM.

kito-cheng:
Thanks for everyone, I gonna merge this :)

riscv-c-api.md
```diff
@@ -235,6 +235,51 @@ long __riscv_clmul (long a, long b); // clmul rd, rs1, rs2
 vint8m1_t __riscv_vadd_vv_i8m1(vint8m1_t vs2, vint8m1_t vs1, size_t vl); // vadd.vv vd, vs2, vs1
 ```
 
+### NTLH Intrisics 
+
+The RISC-V zihintntl extension provides the RISC-V specific intrinsic functions for generating non-temporal memory accesses. These intrinsic functions provide the domain parameter to specify the behavior of memory accesses.
```
topperc, line 240: zihintntl -> Zihintntl

riscv-c-api.md
```diff
@@ -235,6 +235,51 @@ long __riscv_clmul (long a, long b); // clmul rd, rs1, rs2
...
+
+```
+type __riscv_ntl_load (type *ptr, int domain);
+void __riscv_ntl_store (type *ptr, type val, int domain);
+```
+
+There are overloaded functions of `__riscv_ntl_load` and `__riscv_ntl_store`. When these intrinsic functions omit the `domain` argument, the `domain` is implied as `__RISCV_NTLH_ALL`.
+
+```
+type __riscv_ntl_load (type *ptr);
```
topperc, line 255: Is this actually implemented? I don't see a test for it in clang

PR #48, opened 2023-07-26
Function multi-version proposal

BeMg:
During the Function multi-version dispatch the function, we need a method to retrieve the RISC-V hardware environment to make sure all extension must be available.

The problem is

* How to implement this function
* Where to provide this function

From the compiler's view, it will generate the IFUNC resolver when there are more than one implementation with the same symbol name.

Consider following example:

```
__attribute__((target("default"))) int foo (int index)
{
  return index;
}

__attribute__((target("arch=rv64gc"))) int foo (int index)
{
 return index;
}

void bar() {
  foo(0);
}
```

The corresponding assembly will look like:

```
bar() {
(foo.ifunc())(0);
}

.set foo.ifunc, foo.resolver

func_ptr foo.resolver() {
  if (__riscv_ifunc_select("m_a_f_d_c"))
    return ptr foo.m_a_f_d_c;
  return ptr foo.default;
}

int foo.default(int index) {
	...
}

int foo.m_a_f_d_c(int index) {
	...
}
```

The resolver that the compiler generates query and selects for each candidate function. When fulfilling the requirement, then return the corresponding function ptr for further processing.

In this proposal, the major part of the resolver function is `__riscv_ifunc_select`. `__riscv_ifunc_select` must retrieve the hardware information for deciding whether to execute the specific function.

Here we propose that function as the following declaration

```
bool __riscv_ifunc_select(char *FeatureStr);
```

Where FeatureString is a string that concatenates all target features belonging to a particular function. The form can be described in the following BNF form.

When hardware fulfills the FeatureStr, then returns true. Otherwise this function returns false.

---
2023/09/04 Update: The following section take the linux platform as example for `__riscv_ifunc_select` implementation.

There are two ways to retrieve hardware information.

* RISC-V Hardware Probing Interface [1] 
    * Not fully cover all extensions, and need to sync the all defined symbol from linux kernel source code.
* /proc/cpuinfo isa string 
    * Not every system has the cpuinfo file.

Another problem is where to place the function definition.

The compiler-rt/libgcc is a good place to implement these functions, like other target(x86/aarch64) implementation.

[1] https://docs.kernel.org/riscv/hwprobe.html

---
2024/07/11 Update

1. Remain the syntax part, the runtime function move to another RFC.
2. Remove `arch=<full-arch-string>` from syntax

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -319,7 +319,7 @@ EXTENSION-NAME         := Naming rule is defined in RISC-V ISA manual
 For example, the following `foo` function will have three versions but share the same function signature.
 
 ```c
-__attribute__((target_clones("arch=+v", "default", "arch=rv64gc_zbb")))
+__attribute__((target_clones("arch=+v", "default", "arch=+zbb")))
 int foo(int a)
 {
   return a + 5;
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -290,6 +290,104 @@ __attribute__((target("arch=+v"))) int foo(void) { return 0; }
 __attribute__((target("arch=+zbb"))) int foo(void) { return 1; }
 ```
 
+### `__attribute__((target_clones("<TARGET-CLONES-ATTR-STRING>", ...)))
+
+The `target_clones` attribute is used to create multiple versions of a function. The compiler will emit multiple versions based on the provided arguments. 
+
+Each `TARGET-CLONES-ATTR-STRING` defines a distinguished version of the function.  The `TARGET-CLONES-ATTR-STRING` list must include `default` indicating the translation unit scope build attributes.
+
+The syntax of `<TARGET-CLONES-ATTR-STRING>` describes below:
+
+```
+TARGET-CLONES-ATTR-STRING := 'arch=' EXTENSIONS
+                           | 'default'
+
+EXTENSIONS             := <EXTENSION> ',' <EXTENSIONS>
+                        | <EXTENSION>
+
+EXTENSION              := <OP> <EXTENSION-NAME> <VERSION>
+
+OP                     := '+'
+
+VERSION                := [0-9]+ 'p' [0-9]+
+                        | [1-9][0-9]*
+                        |
+
+EXTENSION-NAME         := Naming rule is defined in RISC-V ISA manual
+```
+
+For example, the following `foo` function will have three versions but share the same function signature.
+
+```c
+__attribute__((target_clones("arch=+v", "default", "arch=rv64gc_zbb")))
+int foo(int a)
+{
+  return a + 5;
+}
+
+int bar() {
+  // foo will be resolved by ifunc
+  return foo(1);
+}
+```
+
+It makes the compiler trigger the [function multi-version](#function-multi-version) when there exist more than one version for the same function signature.
+
+### `__attribute__((target_version("<TARGET-VERSION-ATTR-STRING>")))
+
+The `target_version` attribute is used to create one version of a function. Functions with the same signature may exist with multiple versions in the same translation unit.
+
+Each `TARGET-VERSION-ATTR-STRING` defines a distinguished version of the function. If there is more than one version for the same function, it must have `default` one that indicating the translation unit scope build attributes.
+
+The syntax of `<TARGET-VERSION-ATTR-STRING>` describes below:
+
+```
+TARGET-VERSION-ATTR-STRING := 'arch=' EXTENSIONS
+                            | 'default'
+
+EXTENSIONS             := <EXTENSION> ',' <EXTENSIONS>
+                        | <EXTENSION>
+
+EXTENSION              := <OP> <EXTENSION-NAME> <VERSION>
+
+OP                     := '+'
+
+VERSION                := [0-9]+ 'p' [0-9]+
+                        | [1-9][0-9]*
+                        |
+
+EXTENSION-NAME         := Naming rule is defined in RISC-V ISA manual
+```
+
+For example, the following foo function has three versions.
+
+```c
+__attribute__((target_version("arch=+v")))
+int foo(int a)
+{
+  return a + 5;
+}
+
+__attribute__((target_version("arch=+zbb")))
+int foo(int a)
+{
+  return a + 5;
+}
+
+__attribute__((target_version("default")))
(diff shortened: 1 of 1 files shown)


sorear:
This repository is for specifications of features that are portable between multiple RISC-V toolchains.  As such it is inappropriate to specify any behavior exclusively in terms of Linux-only interfaces like hwprobe and cpuinfo.

Providing a string-to-bool or string-to-int (for things like Zicbo* cache block size) lookup interface as a portable frontend to Linux's syscalls, the HWCAP-inspired interfaces on the BSDs, and whatever NT ends up with is a good idea, although it's useful for more than just ifunc; @jrtc27 [suggested](https://libera.irclog.whitequark.org/riscv/2023-07-09#34579173;) `__riscv_get_extension` for essentially this interface.

riscv-c-api.md
```diff
@@ -357,3 +357,51 @@ bar:
...
+
+### Runtime Featrue API
+
+Here is the prototype of that API.
+
+```
+bool __riscv_ifunc_select(char *FeatureString)
+```
+
+Where FeatureString is a string that concatenating all target features belonging to a particular function. The form can be described in the following BNF form.
```
topperc, line 398: concatenating -> concatenates

riscv-c-api.md
```diff
@@ -357,3 +357,51 @@ bar:
 
 The compiler uses an immediate offset of 40 for the `m` constraint, but for the
 `A` constraint uses an extra addi instruction instead.
+
+## Function Multi-version
+
+Function multi-versioning(FMV) provides an approach to selecting the appropriate function according to the runtime environment. This feature is triggered by `target/target_clones` function attribute. The compiler generates the resolver function based on the IFUNC mechanism. It expects that there is an API in the runtime environment for FMV to check if it fulfils all extension requirements.
```
topperc, line 363: fulfils -> fulfills

riscv-c-api.md
```diff
@@ -357,3 +357,51 @@ bar:
...
+
+Here is a example 
+
+| Target Attribute                  | Mangling Name                              |
+|-----------------------------------|--------------------------------------------|
+| target("arch=+v,+zbb,+zicond1p0") | "BASE-FUNC-NAME".experimental-zicond_zbb_v |
+| target("arch=rv64gc")             | "BASE-FUNC-NAME".zifencei_zicsr_m_f_d_c_a  |
+| target("default")                 | "BASE-FUNC-NAME"                           |
+
+NOTE: Should mangling name need to consider the feature come form extension dependency?
```
topperc, line 388: form -> from

riscv-c-api.md
```diff
@@ -357,3 +357,51 @@ bar:
...
+
+### Runtime Featrue API
```
topperc, line 390: Featrue -> Feature

riscv-c-api.md
```diff
@@ -617,3 +655,26 @@ statements, including both RISC-V specific and common operand modifiers.
 | ------------ | --------------------------------------------------------------------------------- | ----------- |
 | z            | Print `zero` (`x0`) register for immediate 0, typically used with constraints `J` |             |
 | i            | Print `i` if corresponding operand is immediate.                                  |             |
+
+## Function Multi-version
+
+Function multi-versioning(FMV) provides an approach to selecting the appropriate function according to the runtime environment. This feature is triggered by `target_version/target_clones` function attribute. The compiler generates the resolver function based on the IFUNC mechanism. It expects that there is an API in the runtime environment for FMV to check if it fulfills all extension requirements.
```
sorear, line 661: Matching ACLE I think this should document the result (it's constant for the life of the process and the first or most specific version) while leaving the rest psABI-defined or implementation-defined, including the runtime feature API.

sorear:
Can `arch=` be removed from target_version and target_clones, since nothing else can appear there? Then it becomes just `[[gnu::target_clones("+zbb","default")]]` or similar.

How are versions and clones prioritized? A big list of every possible exception isn't going to work for us, so it should be something in the source code. Not sure if declaration order would cause problems.

BeMg:
Hi @sorear, thanks for the comment. 

> Can `arch=` be removed from target_version and target_clones, since nothing else can appear there? Then it becomes just `[[gnu::target_clones("+zbb","default")]]` or similar.
> 

For target_version and target_clones's `ATTR-STRING`, I tend to reuse the format like target attribute to avoid confusion. Removing mtune and mcpu could reduce the complexity of usage and could be treated as a subset, but using the format like `[[gnu::target_clones("+zbb","default")]]` is more inconsistent between target attribute and target_clones/target_version. 

From the compiler's perspective, mtune, mcpu information will be highly related to the compilation result. If someday, we need to add mtune and mcpu to target_version /target_clones's `ATTR-STRING`, it will be easier and not break the the existing code.

@kito-cheng any other comments on this topic?

> How are versions and clones prioritized? A big list of every possible exception isn't going to work for us, so it should be something in the source code. Not sure if declaration order would cause problems.

Currently, the selection order depend on IFUNC resolver's implementation. We plan to add the one another option inside `ATTR-STRING` that represents the user's manual priority weight. Like `target_clones("default", "arch=rv64gc;prior=5", "arch=rv64g;prior=7")`.

BeMg:
I have created a pull request https://github.com/llvm/llvm-project/pull/85786 to LLVM to implement `target_clones` in the current proposal. 

I have also created a draft pull request https://github.com/llvm/llvm-project/pull/85790 to implement `__riscv_ifunc_select`, which allows target_clones to run in a QEMU environment. 

For example
```
clang -march=rv64g -rtlib=compiler-rt targetclones.c
qemu-riscv64 -cpu rv64,zbb=true,zba=true -B 0x100000 -L /path/to/sysroot a.out
```

BeMg:
Rebase to origin/main

BeMg:
Update: Remove the arch=<full-arch-string> from syntax

kito-cheng (approved):
LGTM, this has discussed for a while and remove `arch=<full-arch-string>` to prevent some implementation issue is reasonable to me, anyway we gonna merge this and implement that on GCC and LLVM :)

preames:
For anyone following along, there is a follow on proposal to add syntax to assist with selection between multiple candidates, see https://github.com/riscv-non-isa/riscv-c-api-doc/pull/85

PR #49, opened 2023-07-28
Clarify that slow misaligned accesses may be very slow

aswaterman:
The current description might give some people the impression that misaligned accesses are only a little bit slower than aligned ones.  But the intent is for the "slow" macro to be defined when the accesses are legal, but potentially extremely slow.

By contrast, the "fast" macro should be defined when misaligned accesses are reasonably fast, even if they're a little bit slower than misaligned ones.  So, the existing text for the "fast" case is fine.

This isn't really a normative change; it's a tweak to the shades of meaning in the description.


asb (approved):
This phrasing looks like an improvement to me.

kito-cheng (approved):
Thanks:)

PR #52, opened 2023-10-22
Add support for RV64E and LP64E

a4lg:
`__riscv_64e` is defined analogously to `__riscv_32e`.

Changing the meaning of `__riscv_32e` should not cause any problems since if we are just testing for existence of the `E` extension, there is `__riscv_e` preprocessor macro.

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -38,7 +38,8 @@ https://creativecommons.org/licenses/by/4.0/.
 | __riscv             | 1     | Always defined.               |
 | __riscv_xlen        | <ul><li>32 for rv32</li><li>64 for rv64</li><li>128 for rv128</ul> | Always defined.             |
 | __riscv_flen        | <ul><li>32 if the F extension is available **or**</li><li>64 if `D` extension available **or**</li><li>128 if `Q` extension available</li></ul> | `F` extension is available. |
-| __riscv_32e         | 1     | `E` extension is available.   |
+| __riscv_32e         | 1     | RV32E is available.           |
+| __riscv_64e         | 1     | RV64E is available.           |
 | __riscv_vector      | 1     | Implies that any of the vector extensions (`v` or `zve*`) is available |
 | __riscv_v_min_vlen    | <N> (see [__riscv_v_min_vlen](#__riscv_v_min_vlen)) | The `V` extension or one of the `Zve*` extensions is available. |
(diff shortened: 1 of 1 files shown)


cmuellner:
Current status is:

* GCC exposes `__riscv_32e` if `RVE` is enabled (regardless which XLEN is in use). Afaik RVE is limited to rv32 in GCC at the moment, so this change won't break anything.
* LLVM exposes `__riscv_abi_rve` if `ABIName == "ilp32e"` (see `clang/lib/Basic/Targets/RISCV.cpp`).

The referenced GCC patch looks good.
We probably want to have a equal behavior in GCC and LLVM (we could define `__riscv_abi` macros in GCC and define `__riscv_*e` macros in LLVM).

In general, I am fine with the change.

a4lg:
Additional changes:

Added LP64E ABI references to the definitions of `__riscv_abi_rve` and `__riscv_float_abi_soft`.

a4lg:
@cmuellner
Seems GCC also defines `__riscv{,_float}_abi` macros (cf. `riscv_cpu_cpp_builtins` function in `gcc/config/riscv/riscv-c.cc`).

wangpc-pp:
The LLVM implementation (https://reviews.llvm.org/D70401) is the same as the proposal, so I will give a LGTM. Thanks! :-)

cmuellner:
Are there patches for GCC available?

a4lg:
@cmuellner
My RFC PATCH submission:
https://gcc.gnu.org/pipermail/gcc-patches/2023-October/633846.html
and corresponding commit:
https://github.com/a4lg/gcc/commit/72be4bc48fb0a1cdd15a99220fc951ab7aa00d30

I'll rebase against the latest master and make a ping, removing "RFC" status.

cmuellner:
@kito-cheng: any objections to merging this?
The change here is straight-forward and we have patches.
The ABI side (https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/299) is not finalized, but I think this won't have an influence on this PR.

kito-cheng (approved):
We already have `__riscv_e` to check `e` is enabled for both rv64 or rv32, but I think it's OK to add macro for symmetric, so LGTM

a4lg:
Since this proposal and my GCC patch set (with a bit of modification from RFC PATCH) are approved, GCC's RV64E/LP64E support is merged into its master/trunk.
* [The official GCC repository](https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=006e90e13441c3716b40616282b200a0ef689376)
* https://github.com/a4lg/gcc/commit/006e90e13441c3716b40616282b200a0ef689376

PR #55, opened 2023-11-21
Add zicsr and zifencei to extension test macros

hasheddan:
Adds the __riscv_zicsr and __riscv_zifencei extension test macros to the existing table.

cc @kito-cheng


kito-cheng (approved):
it already implement on both open source compiler, and I think explicitly list their should be easier for user who is looking for that pre-defined macro

PR #56, opened 2023-11-28
Added support for zimop extension

JivanH:
Added support for Zimop extension.
For both open-source compilers: GCC and Clang already prepared corresponding patches. Going to publish them.

cc: @kito-cheng

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -510,15 +510,20 @@ The functions are only available if the compiler's `-march` string
 enables the required ISA extension. (Calling functions for not enabled
 ISA extensions will lead to compile-time and/or link-time errors.)
 
+Intrinsics operating on XLEN sized value are not available as there is no type
+defined. If `xlen_t` is added in the future, this can be revisited.
+
 Unsigned types are used as that is the most logical representation for a
 collection of bits.
 
 Sign extension of 32-bit values on RV64 is not reflected in the interface.
 
-| Prototype                                                               | Instruction   | Extension  | Notes          |
-|-------------------------------------------------------------------------|---------------|------------|----------------|
-| `uint32_t __riscv_mopr(uint32_t rs1, const int n);`                     | `mop.r.[n]`   | Zimop      | `n`=[0..31]    |
-| `uint32_t __riscv_moprr(uint32_t rs1, uint32_t rs2, const int n);`      | `mop.rr.[n]`  | Zimop      | `n`=[0..8]     |
+| Prototype                                                               | Instruction   | Extension    | Notes                                                         |
+|-------------------------------------------------------------------------|---------------|--------------|---------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


riscv-c-api.md
```diff
@@ -503,6 +504,22 @@ Sign extension of 32-bit values on RV64 is not reflected in the interface.
...
+
+Unsigned types are used as that is the most logical representation for a
+collection of bits.
+
+Sign extension of 32-bit values on RV64 is not reflected in the interface.
+
+| Prototype                                                               | Instruction   | Extension  | Notes          |
+|-------------------------------------------------------------------------|---------------|------------|----------------|
+| `uint32_t __riscv_mopr(uint32_t rs1, const int n);`                     | `mop.r.[n]`   | Zimop      | `n`=[0..31]    |
+| `uint32_t __riscv_moprr(uint32_t rs1, uint32_t rs2, const int n);`      | `mop.rr.[n]`  | Zimop      | `n`=[0..8]     |
```
kito-cheng, line 521: The interface seems not right to me, it may not work correctly if the address larger than 4G, it will truncated due to the type, so I guess `uintptr_t` may be better type, but maybe uintxlen_t is what we really need here since ... ilp32 for RV64.

cc. @topperc @cmuellner

cmuellner: Or we follow the pattern that @topperc defined for the bitmanip and scalar crypto extensions: use `_32` and `_64` postfixes and restrict the availability of the `_64` API to RV64. Plus the following paragraph:
> Intrinsics operating on XLEN sized value are not available as there is no type defined. If xlen_t is added in the future, this can be revisited.

ln8-8: I noticed that definitions of intrinsic on GCC and LLVM are differ.
For example intrinsic for clmul defined as __riscv_clmul in GCC and __riscv_clmul_32/__riscv_clmul_64 in LLVM.
Should I define with postfixes and implement in each compiler in own way or should implement with the same API?

JivanH: > Or we follow the pattern that @topperc defined for the bitmanip and scalar crypto extensions: use `_32` and `_64` postfixes and restrict the availability of the `_64` API to RV64. Plus the following paragraph:

@ln8-8 updated the patch as you requested, besides he updated patch for LLVM. 
Now it meets these requirements.

cmuellner (approved):
LGTM

cmuellner:
@kito-cheng, any objections?

cmuellner:
Squashed and merged.

Thanks!

topperc:
Since these maybe ops can be used to implement real operations like Zicfiss, I think we need to define some additional behavior.

Can they be reordered with loads/stores? Can they be moved across calls? Can they be speculated? Can they be duplicated?

topperc:
I'm not sure it makes sense to have 32 bit forms of mop intrinsics on RV64? What is the compiler supposed to put in the upper bits? For the bitmanip intrinsics that we did this for, we knew that the upper 32 bits of the input only affected the upper 32 bits of the result or we defined rev8_32 as rev8+sraiw.

For the may-be-ops we don't know what the real instruction will be so we don't know how to treat the upper bits.

jrtc27:
What is the actual use case for these existing, rather than giving named ones for specific ones?

PR #57, opened 2023-11-29
Rework the extension test macros section

cmuellner:
This attempts to rework the extension test macros section and drop the useless (and incomplete) extension test macro list.
Previous attempts to clean up this section can be found in #27 and #28.

This should avoid the need for PRs like #55 that add something obvious done by GCC and LLVM without the need for this list. Further, we recently also discussed this in the context of #44 (adding bitmap and scalar crypto extensions to the list). So I think it is time to remove the list.

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -108,15 +108,25 @@ Perhaps obvious, but these are mutually exclusive, so only one is defined at a t
 for a compilation unit.
 
 
-### Architecture Extension Test Macro
+### Architecture Extension Test Macros
 
-Architecture extension test macro is a new set of test macro to checking the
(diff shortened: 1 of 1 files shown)


cmuellner:
Rebased.

@kito-cheng: ping

asb (approved):
For what it's worth, looks good to me. Definitely better than an incomplete list of extensions.

kito-cheng (approved):
LGTM, apparently I missed this PR somehow, anyway thanks for improving this section, I just got asked around XXX extension has marco to check or not, and every time I point I point to here, I also need spend time to explain the table isn't up to date too, so I think just dropping whole table can make people less confusion (on why XXX extension is not list their) at first time.

PR #58, opened 2023-11-29
Fix the  __riscv_sha512sum0/1 intrinsics for RV32.

topperc:
These should be __riscv_sha512sum0r and __riscv_sha512sum1r. There are only 2 instructions not 4.


cmuellner (approved):
Double-checked against the scalar crypto spec.
LGTM

PR #59, opened 2023-12-05
Fix the __riscv_clmul* intrinsics.

Liaoshihua:
These should be  __riscv_clmul*(uintxlen_t rs1, uintxlen_t rs2). There are only 1 argument.


Liaoshihua:
@topperc @cmuellner

cmuellner (approved):
LGTM

I've also confirmed that these are in sync with the implementation in LLVM.

PR #66, opened 2024-02-01
Document constraint "s"

MaskRay:
"s" can be used to create an artificial reference for linker garbage collection, or define sections to hold symbol addresses. "S" does not work for a preemptible symbol in GCC while "s" does.

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -559,6 +559,7 @@ statements, including both RISC-V specific and common operand constraints.
 | I                  | 12-bit signed immediate integer operand      |      |
 | K                  | 5-bit unsigned immediate integer operand     |      |
 | J                  | Zero integer immediate operand                |      |
+| s                  | symbol or label reference with a constant offset |      |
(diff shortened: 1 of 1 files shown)


cmuellner:
Is there a reference to GCC and/or LLVM patches?

kito-cheng (approved):
LGTM, RISC-V GCC already support that since day 1, and LLVM supporting also under review[1], plan to merge this PR once LLVM supporting merged :)

[1] https://github.com/llvm/llvm-project/pull/80201

kito-cheng:
For GCC: https://github.com/gcc-mirror/gcc/blob/master/gcc/common.md#L97

cmuellner:
I just want to mention that in GCC `s` is an arch-independent constraint: https://gcc.gnu.org/onlinedocs/gcc/Simple-Constraints.html#index-s-in-constraint

But since this won't ever change, it does not matter much.

kito-cheng:
clang support has been merged to trunk :)

PR #67, opened 2024-02-01
Function attribute for standard vector calling convention variant: riscv_vector_cc

kito-cheng:
Standard vector calling convention variant has landed in psABI now, and the implementation also merged into GCC trunk, LLVM part are also under reviewing, and this attribute has mentioned during the discussion, but seems like we don't document that yet.

---

Standard vector calling convention variant will only enabled when function has vector argument or returing value by default, however user may also want to invoke function without that during a vectorized loop at some situation, but it will cause a huge performance penalty due to vector register store/restore.

So user can declare function with this riscv_vector_cc attribute like below, that could enforce function will use standard vector calling convention variant.

```c
void foo() __attribute__((riscv_vector_cc));
[[riscv::vector_cc]] void foo(); // For C++11 and C23
```

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -318,7 +318,14 @@ __attribute__((target("arch=+v"))) int foo(void) { return 0; }
 __attribute__((target("arch=+zbb"))) int foo(void) { return 1; }
 ```
 
-### `__attribute__((riscv_vector_cc))`
+### riscv_vector_cc
+
+Supported Syntaxes:
+| Style  | Syntax                              |
+| ------ | ----------------------------------- |
+| GNU    | `__attribute__((riscv_vector_cc)))` |
+| C++11  | `[[riscv::vector_cc]]`              |
+| C23    | `[[riscv::vector_cc]]`              |
 
 Functions declared with this attribute will use to the standard vector calling
 convention variant as defined in the RISC-V psABI, even if the function has
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -318,6 +318,12 @@ __attribute__((target("arch=+v"))) int foo(void) { return 0; }
 __attribute__((target("arch=+zbb"))) int foo(void) { return 1; }
 ```
 
+### `__attribute__((riscv_vector_cc))`
+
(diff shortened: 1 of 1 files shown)


riscv-c-api.md
```diff
@@ -318,6 +318,12 @@ __attribute__((target("arch=+v"))) int foo(void) { return 0; }
 __attribute__((target("arch=+zbb"))) int foo(void) { return 1; }
 ```
 
+### `__attribute__((riscv_vector_cc))`
+
+Functions declared with this attribute will use to the standard vector calling
+convention variant as defined in the RISC-V psABI, even if the function has
+vector arguments or returing value.
```
cmuellner, line 325: typo: "returing"
Proposal: "...has vector arguments or a return value".

kito-cheng: Oh, damm, it's the only word not reviewed by my grammar checker :P

kito-cheng:
ChangeLog:
- Fix typo

cmuellner (approved):
LGTM
But I haven't checked the compiler side (are there patches or is this upstream?).

sorear:
(copying my comment from #68 since it applies to both)

Stylistically, since this is a well-defined RISC-V specific ABI attribute independent of any GNU extensions, I wonder if it should be `[[riscv:vector_cc]]` instead of `[[gnu:riscv_vector_cc]]`.  Probably not since other architectures still seem to be using attributes in the compiler-specific namespace?

I believe that this needs to be included in C++ function pointer mangling to support natural uses (although this [doesn't work on aarch64](https://gcc.godbolt.org/z/j6aj1E3PY)).  Do the rules for that need to be specified here?

LGTM otherwise.

kito-cheng:
Update:

- Added C++11 and C23 style syntax

kito-cheng:
GCC and clang/LLVM implementation has merged into trunk[1][2]

[1] https://github.com/gcc-mirror/gcc/commit/4547628c78618616595f4b5e1bef2b582c77793d
[2] https://github.com/llvm/llvm-project/pull/77560

PR #68, opened 2024-02-02
Function attribute for standard fixed-length vector calling conventio…

kito-cheng:
…n variant

Fixed-length vector are passed via general purposed register or memory within current ABI design, we proposed a standard fixed-length vector calling convention variant for passing the fixed-length vector via vector register.

This is the syntax part in the proposal, further detail for that calling convention variant see https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/418

### DIFF
--- a/riscv-c-api.md
+++ b/riscv-c-api.md
@@ -318,6 +318,48 @@ __attribute__((target("arch=+v"))) int foo(void) { return 0; }
 __attribute__((target("arch=+zbb"))) int foo(void) { return 1; }
 ```
 
+### `__attribute__((riscv_vls_cc))`
+
+Functions declared with this attribute will utilize the standard fixed-length
+vector calling convention variant instead of the default calling convention
+defined by the ABI. This variant aims to pass fixed-length vectors via vector
+registers, if possible, rather than through general-purpose registers.
+
+The attribute can accept an optional unsigned integer argument within the value
+range of `[32, 65536]`, which must be a power of two. This argument specifies
+the ABI_VLEN. If not provided, the default value is set to 128. However, this
+default value can be changed via command-line options or pragma directives.
+
+```c
+// foo uses the standard fixed-length vector calling convention variant with the
+// default ABI_VLEN, which is 128.
+void foo __attribute__((riscv_vls_cc));
+// bar uses the standard fixed-length vector calling convention variant with
+// ABI_VLEN=256.
+void bar __attribute__((riscv_vls_cc(256)));
+```
+
+One constraint on ABI_VLEN is that it must be larger than or equal to the
+minimal VLEN, as specified by the `-march` option through the `zvl*b` extension,
+pragma directives, or the target attribute.
+
+```c
+// foo is declared to use the standard fixed-length vector calling convention variant
+// with ABI_VLEN=256. Compilation will succeed with -march=rv64gcv_zvl256b, as it
+// supports VLEN of 256. However, compiling with -march=rv64gcv will result in an error,
+// because rv64gcv's VLEN is 128, which is less than the specified ABI_VLEN of 256.
+void foo __attribute__((riscv_vls_cc(256)));
+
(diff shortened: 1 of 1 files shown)


sorear:
Stylistically, since this is a well-defined RISC-V specific ABI attribute independent of any GNU extensions, I wonder if it should be `[[riscv:vls_cc]]` instead of `[[gnu:riscv_vls_cs]]`.  Probably not since other architectures still seem to be using attributes in the compiler-specific namespace?

I believe that this needs to be included in C++ function pointer mangling to support natural uses (although this [doesn't work on aarch64](https://gcc.godbolt.org/z/j6aj1E3PY)).  Do the rules for that need to be specified here?

LGTM otherwise.

kito-cheng:
Update:
- Added C++11 and C23 style syntax

riscv-c-api.md
```diff
@@ -318,6 +318,57 @@ __attribute__((target("arch=+v"))) int foo(void) { return 0; }
...
+
+Functions declared with this attribute will utilize the standard fixed-length
+vector calling convention variant instead of the default calling convention
+defined by the ABI. This variant aims to pass fixed-length vectors via vector
+registers, if possible, rather than through general-purpose registers.
+
+The attribute can accept an optional unsigned integer argument within the value
+range of `[32, 65536]`, which must be a power of two. This argument specifies
+the `ABI_VLEN`. If not provided, the default value is set to 128. However, this
+default value can be changed via command-line options or pragma directives.
```
topperc, line 340: Where are the command line options or pragma defined?

riscv-c-api.md
```diff
@@ -318,6 +318,57 @@ __attribute__((target("arch=+v"))) int foo(void) { return 0; }
...
+
+One constraint on `ABI_VLEN` is that it must be larger than or equal to the
+minimal `VLEN`, as specified by the `-march` option through the `zvl*b` extension,
+pragma directives, or the target attribute.
+
+```c
+// foo is declared to use the standard fixed-length vector calling convention variant
+// with ABI_VLEN=256. Compilation will succeed with -march=rv64gcv_zvl256b, as it
+// supports VLEN of 256. However, compiling with -march=rv64gcv will result in an error,
+// because rv64gcv's VLEN is 128, which is less than the specified ABI_VLEN of 256.
```
topperc, line 359: This disagrees with the preceding paragraph that says ABI_VLEN must be larger or equal than the minimval VLEN.

riscv-c-api.md
```diff
@@ -318,6 +318,57 @@ __attribute__((target("arch=+v"))) int foo(void) { return 0; }
...
+```c
+// foo uses the standard fixed-length vector calling convention variant with the
+// default ABI_VLEN, which is 128.
+void foo __attribute__((riscv_vls_cc));
+// bar uses the standard fixed-length vector calling convention variant with
+// ABI_VLEN=256.
+void bar __attribute__((riscv_vls_cc(256)));
+```
+
+One constraint on `ABI_VLEN` is that it must be larger than or equal to the
```
topperc, line 351: `ABI_VLEN` must be less than or equal to`. No need for the `One constraint`.

riscv-c-api.md
```diff
@@ -318,6 +318,57 @@ __attribute__((target("arch=+v"))) int foo(void) { return 0; }
...
+`*` N is the the `ABI_LEN`.
+
+Functions declared with this attribute will utilize the standard fixed-length
+vector calling convention variant instead of the default calling convention
+defined by the ABI. This variant aims to pass fixed-length vectors via vector
+registers, if possible, rather than through general-purpose registers.
+
+The attribute can accept an optional unsigned integer argument within the value
+range of `[32, 65536]`, which must be a power of two. This argument specifies
+the `ABI_VLEN`. If not provided, the default value is set to 128. However, this
```
topperc, line 339: `is set to` -> `is`

PR #79, opened 2024-06-06
Add atomic intrinsics

wangpc-pp:
We need this when optimizing some locks' implementation, but using inline assembly will stop some optimizations. For example, if we use inline assembly, the compiler doesn't know `lr.w` will sign-extend the result.

ARM provides similar intrinsics like `__builtin_arm_ldaex`.

Currently, we only add intrinsics for `Zawrs`.

### DIFF
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -485,6 +485,17 @@ RISC-V intrinsics examples:
 vint8m1_t __riscv_vadd_vv_i8m1(vint8m1_t vs2, vint8m1_t vs1, size_t vl); // vadd.vv vd, vs2, vs1
 ----
 
+### Atomic Intrinsics
+In order to access the RISC-V atomic intrinsics, it is necessary to include the
+header file `riscv_atomic.h`.
+The functions are only available if the compiler's `-march` string enables the
+required ISA extension.
+The following table contains all the atomic intrinsics:
+| Prototype                 | Instruction | Extension | Notes |
+| ---------                 | ----------- | --------- | ----- |
+| `void __riscv_wrs_nto();` | `wrs.nto`   | Zawrs     | |
+| `void __riscv_wrs_sto();` | `wrs.sto`   | Zawrs     | |
+
 === NTLH Intrinsics
 
 


wangpc-pp:
I don't know if we should add intrinsics for AMO instructions as we have some "standard" C builtins for AMO.

riscv-c-api.md
```diff
@@ -342,6 +342,21 @@ RISC-V intrinsics examples:
...
+
+In order to access the RISC-V atomic intrinsics, it is necessary to include the
+header file `riscv_atomic.h`.
+
+The functions are only available if the compiler's `-march` string enables the
+required ISA extension.
+
+The following table contains all the atomic intrinsics:
+
+| Prototype                                                                   | Instruction        | Extension         | Notes |
```
topperc, line 355: Shrink the width of the first column?

wangpc-pp:
Ping. :-)

riscv-c-api.md
```diff
@@ -342,6 +342,21 @@ RISC-V intrinsics examples:
 vint8m1_t __riscv_vadd_vv_i8m1(vint8m1_t vs2, vint8m1_t vs1, size_t vl); // vadd.vv vd, vs2, vs1
 ```
 
+### Atomic Intrisics
```
topperc, line 345: Intrinsics*

riscv-c-api.md
```diff
@@ -342,6 +342,21 @@ RISC-V intrinsics examples:
...
+required ISA extension.
+
+The following table contains all the atomic intrinsics:
+
+| Prototype                 | Instruction | Extension | Notes |
+| ---------                 | ----------- | --------- | ----- |
+| `void __riscv_wrs_nto();` | `wrs.nto`   | Zawrs     | |
+| `void __riscv_wrs_sto();` | `wrs.sto`   | Zawrs     | |
+
 ### NTLH Intrisics
```
topperc, line 360: It's also mispelled here

wangpc-pp: Fixed in https://github.com/riscv-non-isa/riscv-c-api-doc/pull/83

cmuellner:
@wangpc-pp is there a GCC or LLVM patch that adds support for this?
I see the referenced LLVM PR (which has been closed), which also included LR and SC.

wangpc-pp:
> @wangpc-pp is there a GCC or LLVM patch that adds support for this? I see the referenced LLVM PR (which has been closed), which also included LR and SC.

I almost forgot this PR :-)
If we have the consensus on accepting this, I can create another PR with only `Zawrs`. WDYT? Should I preceed on?

cmuellner:
> I almost forgot this PR :-) If we have the consensus on accepting this, I can create another PR with only `Zawrs`. WDYT? Should I preceed on?

From a comment in the referenced LLVM PR (https://github.com/llvm/llvm-project/pull/94578#issuecomment-2181003807), I can see that this was discussed, and there is acceptance for Zawrs intrinsics, but not for a particular API.

I'm fine with this PR, but we need the OK from the GCC and LLVM community.

I suggest you create an LLVM PR that adds this API (nothing else). Given that this would be a subset of your previous PR, it should be done relatively quickly.

wangpc-pp:
> How are you supposed to use these? WRS refers to the current reservation, but the only way to have control over that is with LR, which means it has to be in assembly already, surely?

I have the question as well so I haven't merged the implementation. Maybe we can provide a higher abstract layer that generates LR/SC+WRS directly?

aswaterman:
We probably don't want to add intrinsics that expose LR and SC individually because of the lack of a forward-progress guarantee for unconstrained LR/SC loops.  Exposing higher-level abstractions that use LR/SC under the hood avoids this pitfall.

This is slightly less of a concern for LR/WRS since there's no functional bug if the reservation yields before the WRS is executed, but it is still not ideal.  Some cores will eagerly yield the reservation as soon as they hit another memory access that isn't an SC, so if the compiler slips a register spill in between the LR and the WRS, such cores won't ever enter a lower-power state.  So I think we want higher-level abstractions in this case, too.

We could define e.g. `wrs_sto_until(addr, value)` to emit something like `1: lr.d t0, (addr); bne t0, value, 1f; wrs.sto; j 1b; 1:`, and `wrs_sto_while` that inverts the branch condition, but unfortunately there might be a proliferation of these for more complex loop-exit conditions.  I guess at some point you just need to write assembly code.

jrtc27:
> This is slightly less of a concern for LR/WRS since there's no functional bug if the reservation yields before the WRS is executed, but it is still not ideal. Some cores will eagerly yield the reservation as soon as they hit another memory access (whether or not it's an SC), so if the compiler slips a register spill in between the LR and the WRS, such cores won't ever enter a lower-power state. So I think we want higher-level abstractions in this case, too.

But if there's an LR that comes in between for whatever reason then you may have a valid reservation for a different location. Waiting with a timeout will then risk waiting a bit too long, but waiting without a timeout will risk waiting indefinitely even though the intended location has been modified. Probably there will be no LR added by the compiler, but who's to say for sure...

aswaterman:
Indeed.  In any case, the upshot is that we shouldn't go down the route of adding intrinsics for the primitives.  Adding intrinsics for entire WRS loops, maybe, if we think we can capture the important use cases without a proliferation of new intrinsics.

wangpc-pp:
> @wangpc-pp, what is the status of this? Did this land in LLVM or GCC?

No, it is not landed and I am not working on this currently. I can close it if it bothers others.

cmuellner:
Thanks for the update!
Let's close this.

PR #82, opened 2024-07-11
Migration to RISC-V specifications template in asciidoc

apazos:


cmuellner (approved):
LGTM

cmuellner:
I've built this PR with `git submodule update --init ; make` on a Fedora machine (using Podman as container runtime) and everything looks ok.

While diffing the original Markdown with the generated PDF, I noticed that there are small differences (e.g. the section scope, the section contributors and the table captions), but nothing that qualifies as regression.

We could tweak the layout a bit (e.g. column widths), but that could be done later on as the tables look not that bad.

That said, I consider this as ready to land.

PR #85, opened 2024-08-07
[FMV] Add `Priority` syntax for version selection order

BeMg:
Function multi-versioning provides several versions to choose from. When selecting the appropriate version, there are two factors that impact this process. First is whether a version is eligible. Second is the order in which all eligible versions are tried. For other targets (like aarch64), a priority number is defined for each feature. In this PR, we propose new syntax that allows users to define their own priority for specific versions.

The version selection process applies the following rules in order:

1. Among the eligible versions, select the one with the highest priority.
2. If multiple versions have equal priority, select the one that was declared first.
3. If no other suitable versions are found, fall back to the "default" version.

### DIFF
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -328,8 +328,21 @@ The syntax of `<TARGET-CLONES-ATTR-STRING>` describes below:
 
 [source, C]
 ----
-TARGET-CLONES-ATTR-STRING := 'arch=' EXTENSIONS
-                           | 'default'
+TARGET-CLONES-ATTR-STRING := 'default'
+                           | ATTR-STRINGS
+
+ATTR-STRINGS              := ATTR-STRING
+                           | ';' ATTR-STRINGS
+
+ATTR-STRING               := ARCH-ATTR ';' PRIORITY-ATTR
+                           | PRIORITY-ATTR ';' ARCH-ATTR
+                           | ARCH-ATTR
+
+ARCH-ATTR                 := 'arch=' EXTENSIONS
+
+PRIORITY-ATTR             := 'priority=' DIGITS
+
+DIGITS                    := [0-9]+
 
 EXTENSIONS             := <EXTENSION> ',' <EXTENSIONS>
                         | <EXTENSION>
@@ -350,7 +363,7 @@ same function signature.
 
 [source, C]
 ----
-__attribute__((target_clones("arch=+v", "default", "arch=+zbb")))
+__attribute__((target_clones("arch=+v;priority=2", "default", "arch=+zbb;priority=1")))
 int foo(int a)
 {
   return a + 5;
@@ -361,6 +374,8 @@ int bar() {
   return foo(1);
 }
 ----
+
+The `priority` accepts a digit as the version priority during [Version Selection](#version-selection). If `priority` isn't specified, then the priority of version defaults to zero.
 
 It makes the compiler trigger the <<function-multi-version, function multi-version>>,
 when there exist more than one version for the same function signature.
@@ -376,38 +391,19 @@ function. If there is more than one version for the same function, it
 must have `default` one that indicating the translation unit scope build
 attributes.
 
-The syntax of `<TARGET-VERSION-ATTR-STRING>` describes below:
-
-[source, C]
-----
-TARGET-VERSION-ATTR-STRING := 'arch=' EXTENSIONS
-                            | 'default'
-
-EXTENSIONS             := <EXTENSION> ',' <EXTENSIONS>
-                        | <EXTENSION>
-
-EXTENSION              := <OP> <EXTENSION-NAME> <VERSION>
-
-OP                     := '+'
-
-VERSION                := [0-9]+ 'p' [0-9]+
-                        | [1-9][0-9]*
-                        |
-
-EXTENSION-NAME         := Naming rule is defined in RISC-V ISA manual
-----
+The syntax of `<TARGET-VERSION-ATTR-STRING>` is the same as described above for `<TARGET-CLONES-ATTR-STRING>`.
 
 For example, the following foo function has three versions.
 
 [source, C]
 ----
-__attribute__((target_version("arch=+v")))
+__attribute__((target_version("arch=+v;priority=1")))
 int foo(int a)
 {
   return a + 5;
 }
 
-__attribute__((target_version("arch=+zbb")))
+__attribute__((target_version("arch=+zbb;priority=2")))
 int foo(int a)
 {
   return a + 5;
@@ -424,6 +420,10 @@ int bar() {
   return foo(1);
 }
 ----
+
+The `priority` accepts a digit as the version priority during [Version Selection](#version-selection). If `priority` isn't specified, then the priority of version defaults to zero.
+
+The `default` version does not accept the priority.
 
 It makes the compiler trigger the <<function-multi-version, function multi-version>>
 when there exist more than one version for the same function signature.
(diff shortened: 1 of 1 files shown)


BeMg:
For example:

```
__attribute__((target_clones("default", "arch=+zba", "arch=+zbb", "arch=+zba,+zbb")))
```

The version `arch=+zba,+zbb` will never be selected in any environment because other eligible versions will always be chosen first.

BeMg:
cc @kito-cheng @topperc

riscv-c-api.md
```diff
@@ -331,6 +337,8 @@ int bar() {
 }
 ```
 
+The `priority` accepts a digit as the version priority during [Version Selection](#version-selection). If `priority` doesn't exist, then the priority of version defaults to zero.
```
topperc, line 340: "doesn't exist" -> "isn't specified"

riscv-c-api.md
```diff
@@ -386,6 +401,8 @@ int bar() {
...
```
topperc, line 404: "doesn't exist" -> "isn't specified"

topperc:
cc: @preames

riscv-c-api.md
```diff
@@ -299,8 +299,12 @@ Each `TARGET-CLONES-ATTR-STRING` defines a distinguished version of the function
 The syntax of `<TARGET-CLONES-ATTR-STRING>` describes below:
 
 ```
-TARGET-CLONES-ATTR-STRING := 'arch=' EXTENSIONS
-                           | 'default'
+TARGET-CLONES-ATTR-STRING := ATTR-STRING
+                            | ';' TARGET-VERSION-ATTR-STRING
```
topperc, line 303: Is this supposed to be TARGET-CLONES-ATTR-STRING?

BeMg:
Note: Update the BNF to make `default` version doesn't accept the other attr-string.

riscv-c-api.md
```diff
@@ -708,3 +734,17 @@ statements, including both RISC-V specific and common operand modifiers.
...
+The process of selecting the appropriate function version during function multi-versioning follows these guidelines:
+
+1. The implementation of the selection algorithm is platform-specific.
+2. Once a version is selected, it remains in use for the entire duration of the process.
+3. Only versions whose required features are all available in the runtime environment are eligible for selection.
+
+The version selection process applies the following rules in order:
+
+1. Among the eligible versions, select the one with the highest priority.
+2. If multiple versions have equal priority, select the one that was declared first.
```
preames, line 749: Do we want to mandate (2)?  We can leave ourselves more flexibility by saying "select one based on an implementation defined heuristic".

riscv-c-api.md
```diff
@@ -299,27 +299,37 @@ Each `TARGET-CLONES-ATTR-STRING` defines a distinguished version of the function
...
-                        | <EXTENSION>
+ATTR-STRINGS              := ATTR-STRING
+                           | ';' ATTR-STRINGS
 
-EXTENSION              := <OP> <EXTENSION-NAME> <VERSION>
+DEFAULT-ATTR-STRING       := 'default'
 
-OP                     := '+'
+ATTR-STRING               := 'arch=' EXTENSIONS
+                           | 'priority=' DIGIT
```
preames, line 311: I don't think this is correct.  Priority without arch isn't allowed per the english description below.

Also, why not priority 15?  Or -1?  In particular, your LLVM patch appears to accept -1.

preames (commented):
Generally supportive of this syntax proposal, subject to inline comments being addressed.

riscv-c-api.md
```diff
@@ -299,27 +299,37 @@ Each `TARGET-CLONES-ATTR-STRING` defines a distinguished version of the function
 The syntax of `<TARGET-CLONES-ATTR-STRING>` describes below:
 
 ```
-TARGET-CLONES-ATTR-STRING := 'arch=' EXTENSIONS
```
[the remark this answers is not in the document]
BeMg: Does it look better now?

riscv-c-api.md
```diff
@@ -708,3 +734,17 @@ statements, including both RISC-V specific and common operand modifiers.
 Function multi-versioning(FMV) provides an approach to selecting the appropriate function according to the runtime environment. The final binary may contain all versions of the function, with the compiler generating all supported versions and the runtime selecting the appropriate one.
 
 This feature is triggered by `target_version/target_clones` function attribute.
+
+### Version Selection
+
+The process of selecting the appropriate function version during function multi-versioning follows these guidelines:
+
+1. The implementation of the selection algorithm is platform-specific.
```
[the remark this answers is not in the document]
BeMg: Updated.

riscv-c-api.md
BeMg: I think it's a good idea to maintain more flexibility. Updated. Thanks!

riscv-c-api.md
BeMg: > I don't think this is correct. Priority without arch isn't allowed per the english description below.

Update BNF to let priority only appear when architecture exists.

> Also, why not priority 15? Or -1? In particular, your LLVM patch appears to accept -1.

Update the DIGIT to allow '-'. Does [0-9]+ already support 15?

riscv-c-api.md
```diff
@@ -342,8 +358,22 @@ Each `TARGET-VERSION-ATTR-STRING` defines a distinguished version of the functio
...
+
+ATTR-STRING                := ARCH-ATTR ';' PRIORITY-ATTR
+                            | PRIORITY-ATTR ';' ARCH-ATTR
+                            | ARCH-ATTR
+
+ARCH-ATTR                  := 'arch=' EXTENSIONS
+
+PRIORITY-ATTR              := 'priority=' DIGIT
+
+DIGIT                     := '-' [0-9]+
```
preames, line 375: Naming: DIGITS (with an S), or NUMBER.

riscv-c-api.md
```diff
@@ -342,8 +358,22 @@ Each `TARGET-VERSION-ATTR-STRING` defines a distinguished version of the functio
 The syntax of `<TARGET-VERSION-ATTR-STRING>` describes below:
 
 ```
-TARGET-VERSION-ATTR-STRING := 'arch=' EXTENSIONS
-                            | 'default'
+TARGET-VERSION-ATTR-STRING := 'default'
```
preames, line 361: You have this same non-trivial BNF repeated twice.  Maybe replace the second with a back reference to the previous?  "is the same as described above for target_clones"?  (You can also link that text.)

riscv-c-api.md
preames: "DIGIT" is not plural which is what caused my misreading.  :)

riscv-c-api.md
```diff
@@ -331,6 +345,8 @@ int bar() {
 }
 ```
 
+The `priority` accepts a digit as the version priority during [Version Selection](#version-selection). If `priority` isn't specified, then the priority of version defaults to zero.
```
preames, line 348: Now that we accept -1, we have to state ordering.  Does "-1" compare less than "1" (i.e. signed ordering), or does it mean "maximum priority" (i.e. unsigned ordering)?  I think we probably want unsigned here?

BeMg: I assume the signed ordering here, and I also use it in the LLVM implementation. It's just more intuitive to me, but I can change to unsigned ordering if necessary.

topperc: Why do we need to allow signed numbers at all?

topperc: @preames do you see any reason to support negative numbers here? The number of target_clones/target_version attributes should be small enough that setting even a smallish number like 1000 should be high enough to be maximum priority.

preames: Nope, restricting it to positive only would be fine.  I just want to make sure the interpretation is unambiguous.

BeMg: Removing negative numbers from the priority appears to be a clearer solution to avoid ambiguity. I will update this in the pull request.

cyyself:
There hasn't been any update for a few months.

Since this feature has been implemented in both GCC and LLVM, should it be updated to resolve the conflicts and merge?

kito-cheng:
> should it be updated to resolve the conflicts and merge?

I think so, @BeMg could you update this?

kito-cheng (approved):
LGTM and I believe this already included on both GCC and LLVM trunk now

PR #86, opened 2024-09-10
Add __riscv_cmodel_large define for large code model

asb:
With https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/388 landed it makes sense to have a define for the large code model for consistency with medany and medlow.


lenary (approved):
The obvious change - looks good to me.

kito-cheng (approved):

PR #87, opened 2024-09-12
Add macro to test endianness

ismaeldamiao:
The RISC-V ELF Specification allow both big-endian and little-endian but there is no macro on RISC-V C API to check the host endianness.

### DIFF
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -13,6 +13,12 @@ a|
 * 128 for rv128
 |Always defined.
 
+|`+__riscv_endianness+`
+a|
+* 'B' for big-endian
+* 'L' for little-endian
+|Always defined.
+
 |`+__riscv_flen+`
 a|
 * 32 if the F extension is available *or*


cmuellner (commented):
Why not using the `__BYTE_ORDER__` macros that are available on GCC and LLVM?
E.g.:
```
#if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
    ...
#else
   ...
#endif
```

ismaeldamiao:
Simple, because this is a compiler dependent feature, but is reasonable to suppose that any compiler targeting RISC-V have support to RISC-V C API language extensions.

aswaterman:
This is probably a solution in search of a problem, but couldn't we just mandate that, as part of the RISC-V C API, the `__BYTE_ORDER__` etc. macros be defined?

It reflects reality, after all.

ismaeldamiao:
In fact, looks better.

PR #89, opened 2024-10-04
Fix typo zawra should be zawrs on c-api.adoc

cyyself:


cyyself:
Fixed at bfe4aa0d0c1ccdf3c524b172ef31897099d70d27

PR #92, opened 2024-10-15
Inline Asm Constraints and Modifiers - RVC, Raw Encodings, Pairs

lenary:
We have customers with usecases that want more kinds of register
constraints and modifiers. This change proposes support for these
constraints and modifiers, and their names.

Broadly, these are intended to make it easier for users who want to
manually assemble instructions inside inline assembly blocks, either
using the existing instruction formats, or using the raw form of the
`.insn` directive. This makes it easier for hardware designers to
experiment on new ISA extensions, and makes it easier to support
the use of proprietary extensions with unmodified open-source
toolchains.

There are three groups of additions here:

- Constraints for RVC-compatible registers. These use the `c` prefix on
  an existing register constraint, so `cr` gives a GPR between x8-x15,
  and `cf` does the same for an FPR between f8-f15.

  I'm not aware of compressed vector instructions, but we could add
  `cvr`, `cvd` and `cvm` in the future if the core architecture ends up
  having the concept of a vector register with an RVC encoding.

- A modifier, `N`, to print the raw encoding of a register. This is used
  when using `.insn <length>, <encoding>`, where the user wants to pass
  a value to the instruction in a known register, but where the
  instruction doesn't follow the existing instruction formats, so the
  assembly parser is not expecting a register name, just a raw integer.

- Constraints for even-odd pairs of general-purpose registers. These use
  the `R` constraint.

  While the concept of even-odd register pairs is reasonably "new",
  there are places in the architecture where these already exist - the
  doubleword/quad CAS in Zacas, and they are also present in the Zilsd
  specification.

---

There's also a commit which fixes a header for the Assembly Constraints table.

### DIFF
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -752,7 +752,7 @@ to aid portability of floating-point code.
 |J                  |Zero integer immediate operand               |
 |s                  |symbol or label reference with a constant offset |
 |cr                 |RVC general purpose register (`x8`-`x15`)  |
-|cf                 |RVC floating-point register (`f8`-`f15`)   |
+|cf                 |RVC floating-point register (`f8`-`f15` or `x8-x15` with `Zfinx`) |
 |Pr                 |Even-odd general purpose register pair     |
 |vr                 |Vector register                    |
 |vd                 |Vector register, excluding v0      |
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -733,6 +733,11 @@ Sign extension of 32-bit values on RV64 is not reflected in the interface.
 This section lists operand constraints that can be used with inline assembly
 statements, including both RISC-V specific and common operand constraints.
 
+"Floating-point register" in both the `f` and `cf` rows means "a register
+suitable for passing a floating-point value", so when using the `Zfinx`,
+`Zdinx`, or `Zhinxmin` extensions this will allocate an X register. This is done
+to aid portability of floating-point code.
+
 .Constraints on Operands of Inline Assembly Statements
 [%autowidth]
 |===
@@ -747,12 +752,18 @@ statements, including both RISC-V specific and common operand constraints.
 |J                  |Zero integer immediate operand               |
 |s                  |symbol or label reference with a constant offset |
 |cr                 |RVC general purpose register (`x8`-`x15`)  |
-|cf                 |RVC floating point register (`f8`-`f15`)   |
+|cf                 |RVC floating-point register (`f8`-`f15`)   |
 |Pr                 |Even-odd general purpose register pair     |
 |vr                 |Vector register                    |
 |vd                 |Vector register, excluding v0      |
 |vm                 |Vector register, only v0           |
 |===
+
+The `Pr` constraint should print as the even register in the pair, as this
+matches how the `amocas.q` instruction (on RV64) or the `amocas.d` and `Zdinx`
+instructions (on RV32) expect to parse their pair register operands. However,
+both registers in the pair should be considered to be live or clobbered
+together.
 
 NOTE: Immediate value must be a compile-time constant.
 
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -746,12 +746,17 @@ statements, including both RISC-V specific and common operand constraints.
 |K                  |5-bit unsigned immediate integer operand     |
 |J                  |Zero integer immediate operand               |
 |s                  |symbol or label reference with a constant offset |
+|cr                 |RVC general purpose register (`x8`-`x15`)  |
+|cf                 |RVC floating point register (`f8`-`f15`)   |
+|Pr                 |Even-odd general purpose register pair     |
 |vr                 |Vector register                    |
 |vd                 |Vector register, excluding v0      |
 |vm                 |Vector register, only v0           |
 |===
 
 NOTE: Immediate value must be a compile-time constant.
+
+NOTE: The `c*` and `P*` constraints are designed to be extensible to more kinds of registers in the future.
 
 === The Difference Between `m` and `A` Constraints
 
@@ -809,6 +814,7 @@ statements, including both RISC-V specific and common operand modifiers.
 |*Modifiers*       |*Description*                                                                  |*Note*
 |z            |Print `zero` (`x0`) register for immediate 0, typically used with constraints `J` |
 |i            |Print `i` if corresponding operand is immediate.                                  |
+|N            |Print register encoding as integer (0-31). |
 |===
 
 [id=function-multi-version]
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -736,7 +736,7 @@ statements, including both RISC-V specific and common operand constraints.
 .Constraints on Operands of Inline Assembly Statements
 [%autowidth]
 |===
-|*Constraint*       |                                             |*Note*
+|*Constraint*       |*Description*                                |*Note*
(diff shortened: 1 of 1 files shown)


src/c-api.adoc
```diff
@@ -746,13 +746,18 @@ statements, including both RISC-V specific and common operand constraints.
 |K                  |5-bit unsigned immediate integer operand     |
 |J                  |Zero integer immediate operand               |
 |s                  |symbol or label reference with a constant offset |
+|cr                 |RVC general purpose register (`x8`-`x15`)  |
+|cf                 |RVC floating point register (`f8`-`f15`)   |
+|Pr                 |Even-odd general purpose register pair     |
```
topperc, line 751: Do any other targets support pair? I haven't been able to find any in LLVM and I don't know how to implement it in LLVM.

lenary: AArch64 supports tuples of 8 64-bit registers, as implemented here https://github.com/llvm/llvm-project/commit/7d940432c46be83b8fcb5dbefee439585fa820cd

I noticed that the NXP LLVM fork with Zilsd support seems to allocate both `v2i<xlen>` and `i<2*xlen>` to the paired register class, but maybe we only need to do the former? I'm not sure, as it would be nice to be able to pass `uint<2*xlen>_t` to a GPR pair constraint.

I understand that pair constraint is the most complex part of this proposal. I have most of a patch for the other constraints, which I will try to finish in the next few days and post to LLVM.

lenary:
I have an LLVM implementation of the `c*` constraints and the `N` modifier here https://github.com/llvm/llvm-project/pull/112561 which has been approved.

I am still working on the implementation of the `Pr` constraint.

lenary:
There was a question on #39 about whether the paired constraint means "an even GPR", or a "even-odd GPR pair", as both would likely print exactly the same as the even subregister. 

This matters for liveness analysis as, if `Pr` meant "an even GPR", the odd GPR in the pair might be allocated to a different input/output for the inline assembly block, which would conflict with the use of the even/odd value for a pair.

So, to be clear, `Pr` means "treat this as an even-odd GPR pair for liveness" and "print this as the even subregister", so if the compiler choses `a0/a1`, then `a0` is printed, but `a1` will also be defined/used by the inline assembly block.

lenary:
I just opened https://github.com/llvm/llvm-project/pull/112983 which implements the `Pr` constraint.

lenary:
Gentle Ping

Is there anything missing from the textual additions that is needed to clarify the proposal?

How are people feeling about this proposal?

src/c-api.adoc
lenary: I looked this up later, and on LLVM, SystemZ also supports GPR pairs using `r` - https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/SystemZ/SystemZISelLowering.h#L444-L446

kito-cheng:
The proposal look good, let me make sure it's implementable on GCC side before we moving forward...:P

src/c-api.adoc
```diff
@@ -746,13 +746,18 @@ statements, including both RISC-V specific and common operand constraints.
 |K                  |5-bit unsigned immediate integer operand     |
 |J                  |Zero integer immediate operand               |
 |s                  |symbol or label reference with a constant offset |
+|cr                 |RVC general purpose register (`x8`-`x15`)  |
+|cf                 |RVC floating point register (`f8`-`f15`)   |
```
kito-cheng, line 750: Should we treat this as `cr` (`x8`-`x15`) for `Zfinx` or just reject that if `Zfinx`?

I prefer treat treat as `cr` for `Zfinx` so that user can  easier to write `Zfinx`-portable code

lenary: Turns out that Craig already implemented the latter in LLVM: https://github.com/llvm/llvm-project/commit/1bc1a79a65a93a0224b5e5f69584219f9981bd23 (which I am happy with the direction of)

I'm going to push an update noting that "floating point register" means a register suitable for passing a floating point value, so with the `Z*inx` architectures, this means a GPR.

src/c-api.adoc
lenary: I went looking at this and honestly, using a specific constraint rather than something implicit in a modifier seems much more obvious. But thanks for pointing out they had it, the LLVM implementation (they do a bunch of fixing things up later in isel rather than in lowering) was so strange that I hadn't come across what was going on. The RISC-V implementation has ended up following what SystemZ did, which has some overlaps with how Arm works but not many.

kito-cheng: GCC can't use either `P` or `p` as prefix for constraint due to implementation issue, `P` can only used for integer immediate value only, and `p` can only used for holding address value.

I tried to remove that limitation but I found there is several internal checking are rely on the rule...so I propose "RPr" instead of "Pr"...

NOTE: I have primary implementation which cover all of this PR except using "RPr" rather than "Pr".

[1] https://gcc.gnu.org/onlinedocs/gccint/Simple-Constraints.html

kito-cheng: BTW, GCC must use constraint to implement register pair, modifier is impossible to implement on GCC side.

lenary: I think `RPr` is a bit of a mouthful, when most constraints are one letter (and I haven't entirely got my head around the fact that one can provide multiple alternative constraints for the compiler to choose). How about your original suggestion of `R` - this is still "extensible" in that we should be able to use `F` for FPR pairs if we ever need to, and things like `cR` if we ever feel we need to.

I can update the wording here, and updating my LLVM PR won't take too long.

kito-cheng: 'F' is not available on GCC side, it reserved for immediate floating operand...

kito-cheng: Also one more constraint on GCC side, a multi-letter constraint must have same length, e.g. define both `Rp` and `Rpr` are invalid, since it come with length 2 and 3.

lenary: I'm starting to wonder how extensible we really need pairs to be. So far they've only turned up in the ISA for GPRs. I think for vector tuples, i think right now it's possible to just use `vr` or `vd` with the right type of argument?

So maybe all we need is GPR pairs, and we don't need to think about extensibility?

lenary: I've pushed an update with `R`, just saying that `c*` is extensible and pairs are not. We can still change the letter though.

kito-cheng: > i think right now it's possible to just use `vr` or `vd` with the right type of argument?

Yes, just `vr` or `vd` for vector tuple is fine.

kito-cheng: > So maybe all we need is GPR pairs, and we don't need to think about extensibility?

I am OK with `R` for GPR pair only for now :)

kito-cheng (approved):
LGTM

cmuellner:
Do we have a GCC patch for this?

cmuellner (approved):

kito-cheng:
Just let you know the reason why I didn't send the GCC patch yet: Our server are under relocation and I forgot to push that to github...so I need to wait few more day until the server up again...

kito-cheng:
We have both GCC and LLVM, so I gonna merge this PR :)

GCC Part: https://patchwork.sourceware.org/project/gcc/list/?series=41826, this will merge once CI passed.
LLVM Part: https://github.com/llvm/llvm-project/pull/112983, https://github.com/llvm/llvm-project/pull/112561

PR #93, opened 2024-10-25
Add CMO extension intrinsics

Liaoshihua:
Add CMO extension intrinsics description.

### DIFF
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -485,6 +485,48 @@ RISC-V intrinsics examples:
 vint8m1_t __riscv_vadd_vv_i8m1(vint8m1_t vs2, vint8m1_t vs1, size_t vl); // vadd.vv vd, vs2, vs1
 ----
 
+=== CMO Intrinsics
+
+The RISC-V CMO extension provides instruction for perform operations on copies
+of data in the memory hierarchy.
+
+In order to access the RISC-V CMO intrinsics, it is necessary to
+include the header file `riscv_cmo.h`.
+
+The functions are only available if the compiler's `-march` string enables the
+required ISA extension.
+
+.CMO Extension Intrinsics
+[%autowidth]
+|===
+|*Prototype*                                                                              |*Instruction*        |*Extension*    |*Notes*
+|`+void __riscv_cmo_clean(void *addr);+`                                                  |`cbo.clean`          |Zicbom         |
+|`+void __riscv_cmo_flush(void *addr);+`                                                  |`cbo.flush`          |Zicbom         |
+|`+void __riscv_cmo_inval(void *addr);+`                                                  |`cbo.inval`          |Zicbom         |
+|`+void __riscv_cmo_zero(void *addr);+`                                                   |`cbo.zero`           |Zicboz         |
+|`+void __riscv_cmo_prefetch(void *addr, const int rw, const int locality);+`             |`prefetch.[r][w]`    |Zicbop         | `rw`= [0,1], `locality` = [0..3].
+|`+int  __riscv_cmo_prefetchi(const int locality);+`                                      |`prefetch.i`         |Zicbop         |
+|===
+
+Note: riscv_cmo_prefetch is a wrapper around builtin_prefetch.
+Similar to Prefetch Intrinsics, the following table presents the mapping from
+the riscv_cmo_prefetch function to the corresponding assembly instructions,
+assuming the presence of the Zihintntl and Zicbop extensions.
+
+.CMO Functions to Assembly Mapping with Zihintntl
+[%autowidth]
+|===
+|*Prefetch function*                                 |*Assembly*
+|`+__riscv_cmo_prefetch(ptr, 0, 0 /* locality */);+` |`ntl.all + prefetch.r (ptr)`
+|`+__riscv_cmo_prefetch(ptr, 0, 1 /* locality */);+` |`ntl.pall + prefetch.r (ptr)`
+|`+__riscv_cmo_prefetch(ptr, 0, 2 /* locality */);+` |`ntl.p1 + prefetch.r (ptr)`
+|`+__riscv_cmo_prefetch(ptr, 0, 3 /* locality */);+` |`prefetch.r (ptr)`
+|`+__riscv_cmo_prefetch(ptr, 1, 0 /* locality */);+` |`ntl.all + prefetch.w (ptr)`
+|`+__riscv_cmo_prefetch(ptr, 1, 1 /* locality */);+` |`ntl.pall + prefetch.w (ptr)`
+|`+__riscv_cmo_prefetch(ptr, 1, 2 /* locality */);+` |`ntl.p1 + prefetch.w (ptr)`
+|`+__riscv_cmo_prefetch(ptr, 1, 3 /* locality */);+` |`prefetch.w (ptr)`
+|===
+
 === NTLH Intrinsics
 
 


src/c-api.adoc
```diff
@@ -485,6 +485,29 @@ RISC-V intrinsics examples:
...
+
+.CMO Extension Intrinsics
+[%autowidth]
+|===
+|*Prototype*                                                                              |*Instruction*        |*Extension*
+|`+void __riscv_cbo_clean(void *addr);+`                                                  |`cbo.clean`          |Zicbom
+|`+void __riscv_cbo_flush(void *addr);+`                                                  |`cbo.flush`          |Zicbom
+|`+void __riscv_cbo_inval(void *addr);+`                                                  |`cbo.inval`          |Zicbom
+|`+void __riscv_cbo_zero(void *addr);+`                                                   |`cbo.zero`           |Zicboz
+|`+void __riscv_cbo_prefetch(void *addr, const int rw, const int locality);+`             |`prefetch.[r][w]`    |Zicbop
```
topperc, line 507: Is `__riscv_cbo_prefetch` a wrapper around `__builtin_prefetch` where there are 4 possible localities? RISC-V supports 5 localities, 4 NTL hint prefixes and 1 without any prefix.

yulong18:
I follow the documentation to set the __riscv_cbo_prefetchi parameter type, but I get such an error that is **invalid argument to built-in function**. So, I think the authour need to redefine the parameter type.

topperc:
> I follow the documentation to set the __riscv_cbo_prefetchi parameter type, but I get such an error that is **invalid argument to built-in function**. So, I think the authour need to redefine the parameter type.
> 
> 

The rw and locality argument to __builtin_prefetch must be compile time constants. You'll need to use a macro instead of a function.

src/c-api.adoc
Liaoshihua: Yes, this is __builtin_prefetch's wrapper. The 4 NTL hint maping had been included in Prefetch Intrinsics part on this doc. So, this commit only support a wrapper for __builtin_prefetch on zicbop extension.

topperc: Can you document how the different rw/locality map to different instruction sequences?

Liaoshihua: > Can you document how the different rw/locality map to different instruction sequences?

Do you mean to expand to describe these mappings? 
Ok, I will send a new version tomorrow.

src/c-api.adoc
```diff
@@ -485,6 +485,38 @@ RISC-V intrinsics examples:
...
+|`+void __riscv_cbo_prefetch(void *addr, const int rw, const int locality);+`             |`prefetch.[r][w]`    |Zicbop         | `rw`= [0,1], `locality` = [0..4].
+|`+int  __riscv_cbo_prefetchi(const int locality);+`                                      |`prefetch.i`         |Zicbop         |
+|===
+
+Note: riscv_cbo_prefetch is a wrapper around of builtin_prefetch
+[source, C]
+----
+void __riscv_cbo_prefetch(void *addr, const int rw, const int locality)
+{
+  __builtin_prefetch(addr, rw, locality);
```
topperc, line 516: We need to explain that locality==0, adds `nt.all` hint. locality==1, adds `ntl.pall` hint. locality==2, adds `ntl.p1` hint.

src/c-api.adoc
```diff
@@ -485,6 +485,48 @@ RISC-V intrinsics examples:
...
+
+.CMO Extension Intrinsics
+[%autowidth]
+|===
+|*Prototype*                                                                              |*Instruction*        |*Extension*    |*Notes*
+|`+void __riscv_cbo_clean(void *addr);+`                                                  |`cbo.clean`          |Zicbom         |
+|`+void __riscv_cbo_flush(void *addr);+`                                                  |`cbo.flush`          |Zicbom         |
+|`+void __riscv_cbo_inval(void *addr);+`                                                  |`cbo.inval`          |Zicbom         |
+|`+void __riscv_cbo_zero(void *addr);+`                                                   |`cbo.zero`           |Zicboz         |
+|`+void __riscv_cbo_prefetch(void *addr, const int rw, const int locality);+`             |`prefetch.[r][w]`    |Zicbop         | `rw`= [0,1], `locality` = [0..4].
```
topperc, line 507: locality is only [0..3] right?

src/c-api.adoc
```diff
@@ -485,6 +485,48 @@ RISC-V intrinsics examples:
...
+include the header file `riscv_cmo.h`.
+
+The functions are only available if the compiler's `-march` string enables the
+required ISA extension.
+
+.CMO Extension Intrinsics
+[%autowidth]
+|===
+|*Prototype*                                                                              |*Instruction*        |*Extension*    |*Notes*
+|`+void __riscv_cbo_clean(void *addr);+`                                                  |`cbo.clean`          |Zicbom         |
```
topperc, line 503: The patch that landed in gcc uses __riscv_cmo_clean?

cmuellner:
So, we have the situation that the intrinsics landed in GCC about two months ago (d2c8548e0ce51dac6bc51d37236c50f98fca82f0). But we still have valid comments here, which we should take care of.
The good thing is that the recent PR update brought this PR almost into sync with the GCC implementation.

Before going deeper into the details, I'd like to mention that GCC has a generic builtin for prefetching, which also supports locality information in the form of a compile-time constant in the range of zero to three (see https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html). A mapping from these arch-independent locality levels to NTL instructions was done in summer 2024 (bf26413fc4081dfd18b915580b35bdb71481327e): `0 -> ntl.all`, `1 -> ntl.pall`, `2 -> ntl.p1`.

The current GCC implementation of `__riscv_cmo_prefetch()` just invokes the arch-independet `__builtin_prefetch()` with the same arguments. This means that GCC does currently not support `ntl.s1`. Further, GCC does not support emitting a prefetch instruction without a preceding NTL instruction if Zihintntl is enabled. In case Zihintntl is not enabled, the locality information is ignored.

I don't think a 1:1 mapping from GCC arch-independent locality encodings to `__riscv_cmo_prefetch()` is necessary. Instead, I propose the following encoding:
```
#define __riscv_ntl_undefined 0
#define __riscv_ntl_p1 1
#define __riscv_ntl_pall 2
#define __riscv_ntl_s1 3
#define __riscv_ntl_all 4
```

Other compile-time constants for the locality parameter of `__riscv_cmo_prefetch()` should trigger a compile-time error. The behavior of `__riscv_cmo_prefetch()` could be the same regardless of whether Zihintntl is enabled or not. I.e., if the locality parameter is not zero, we can always emit the hint instruction.

Let me know what you think.

Liaoshihua:
I wrote this proposal with the intention of addressing: https://github.com/riscv-non-isa/riscv-c-api-doc/issues/88，

> see https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html)

I noticed that GCC has __builtin_prefetch, so __riscv_cmo_prefetch may not be necessary.

> #define __riscv_ntl_undefined 0
> #define __riscv_ntl_p1 1
> #define __riscv_ntl_pall 2
> #define __riscv_ntl_s1 3
> #define __riscv_ntl_all 4

The following are included in https://github.com/riscv-non-isa/riscv-c-api-doc/blob/fbf5f4ef57c0a239af30f457683830e56cd23b33/src/c-api.adoc?plain=1#L488

So, I would remove prefetch in this PR and only include __riscv_cmo_prefetchi, __riscv_cmo_clean, __riscv_cmo_inval, __riscv_cmo_flush, and __riscv_cmo_zero in riscv_cmo.h

Issue #96, opened 2024-11-14
[FMV] Expilicit allows different vendor share the bit defination of __riscv_vendor_feature_bits


cyyself:
In PR #74, we design [vendor extension is guarded by vendorID](https://github.com/riscv-non-isa/riscv-c-api-doc/pull/74#issuecomment-2128844747). However, here we missed some cases like:

1. An SoC vendor buys a CPU IP from another CPU IP vendor with some modification to CPU RTL and then fills marchid with itself mvendorid
2. An SoC vendor uses multiple CPU IPs from different vendors for their SoC and customizes the CPU IP to have the same vendor extension, which multiple CPU IPs have different mvendorid

These cases are discussed from [Linux Kernel mailing list on hwprobe support for T-HEAD vendor extensions thread](https://lore.kernel.org/linux-riscv/20240416-husband-flavored-96c1dad58b6e@wendy/). I think their concerns are right.

So we might need to make the `__riscv_vendor_feature_bits` not guarded by `mvendorid`. Since this hasn't been specified in the c-api-doc currently, it's not too late to make this change.

cyyself:
cc @kito-cheng @BeMg @ConchuOD @charlie-rivos @evangreen

jrtc27:
I agree the current interface is insufficient. One option would be to have __riscv_vendor_feature_bits be a list of (vendor, pointer) pairs, where the pointer points to an individual instance of what __riscv_vendor_feature_bits is today (i.e. the per-vendor bits). But for now we should just remove __riscv_vendor_feature_bits from the spec now and come up with a fixed solution in future.

cyyself:
> Yeah, I agree to remove that for now since we don't have any vendor bits yet. so maybe we postpone that a little bit is best way.
> 
> [@cyyself](https://github.com/cyyself) could you create a PR to drop that? and then we could have few more time to discuss the mechanism.

OK. But it’s too late at UTC+8. Maybe I will do that tomorrow.

cyyself:
> One option would be to have __riscv_vendor_feature_bits be a list of (vendor, pointer) pairs, where the pointer points to an individual instance of what __riscv_vendor_feature_bits is today (i.e. the per-vendor bits). 

[...]

PR #101, opened 2025-01-17
Drop __riscv_vendor_feature_bits

cyyself:
As discussed in #96, current interface is insufficient to support some cases, like a vendor buying a CPU IP from the upstream vendor but using their own mvendorid and custom features from the upstream vendor. In this case, we might need to add these extensions for each downstream vendor many times. Thus, making __riscv_vendor_feature_bits guarded by mvendorid is not a good idea. So, drop __riscv_vendor_feature_bits for now, and we should have time to discuss a better solution.

### DIFF
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -839,7 +839,7 @@ This feature is activated by the `target_version/target_clones` function attribu
 
 === Extension Bitmask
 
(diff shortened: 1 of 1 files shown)


kito-cheng:
Both LLVM and GCC has merged:
https://github.com/gcc-mirror/gcc/commit/2605daa6b896aed15dead194462725874f332c0a
https://github.com/llvm/llvm-project/commit/2cd8207b26ea4269630feba661f68554d7ae3c15

cmuellner (approved):

PR #102, opened 2025-01-23
Add RVC GPR Pair Constraint - `cR`

lenary:
This follows the notion of using `c` as a prefix to mean "RVC-compatible", and `R` to mean "even-odd GPR Pair".

The Zclsd specification adds registers with this constraint, in particular the `c.ld` and `c.sd` instructions need it.

### DIFF
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -755,6 +755,7 @@ to aid portability of floating-point code.
 |cr                 |RVC general purpose register (`x8`-`x15`)  |
 |cf                 |RVC floating-point register (`f8`-`f15` or `x8-x15` with `Zfinx`) |
 |R                  |Even-odd general purpose register pair     |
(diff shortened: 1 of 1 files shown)


kito-cheng (approved):
LGTM, it's reasonable and also trivial to implement, I saw you already implemented in LLVM and I believe it's also trivial to support on GCC side, although it's kinda late to GCC since it's become bug fix only development stage, so that may defer to GCC 16, but I think it's fine to moving forward here.

lenary:
Sorry for missing your release window :( I figured this would be uncontroversial enough that I could add it to LLVM for our release.

kito-cheng:
The patch has been sent, and will land once CI pass: https://gcc.gnu.org/pipermail/gcc-patches/2025-May/683737.html

lenary:
Thanks @kito-cheng!

PR #106, opened 2025-05-12
Remove User Interrupts, Relax Requirements

lenary:
These have been removed from both GCC and Clang, because the underlying N extension they were based on was never ratified.

This change also relaxes some of the requirements, to allow vendor interrupt modes, and combinations of (compatible) interrupt modes, which matches the LLVM implementation of SiFive interrupts.

### DIFF
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -183,7 +183,7 @@ This attribute is incompatible with the `interrupt` attribute.
 NOTE: Be aware that compilers might have further restrictions on naked
 functions. Please consult your compiler's manual for more information.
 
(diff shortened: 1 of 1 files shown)


topperc (approved):
LGTM, but wait for @kito-cheng

kito-cheng (approved):

cmuellner (approved):

cmuellner:
@lenary has highlighted that GCC still supports user-level interrupt handlers (see https://github.com/riscv-non-isa/riscv-c-api-doc/pull/116#issuecomment-3115495520).

I created a GCC patch for this (https://gcc.gnu.org/pipermail/gcc-patches/2025-July/690574.html).

PR #107, opened 2025-05-13
Add bitmask for Q

el-ev:
Q is the standard extension for Quad-precision floating point.
https://github.com/llvm/llvm-project/pull/139369 is going to add assembly support for the Q extension.

https://github.com/riscv/riscv-isa-manual/blob/main/src/machine.adoc


topperc:
binutils already has assembler support. Why does LLVM adding it require a new feature bit if binutils didn't?

cmuellner (approved):

PR #108, opened 2025-05-15
Add E and Q ext to Extension Bitmask

Liaoshihua:
Embedded and Quad-precision floating-point extensions' bitmask have been defined in misa bit position. The implement of Q ext in GCC needs the bitmask. 

In addition, the bitmask of zqinx、zdinx、zfinx、zhinx have not been defined. Should they be added at the end of bitmask?


Liaoshihua:
I didn't notice that there was already https://github.com/riscv-non-isa/riscv-c-api-doc/pull/107 . Close this PR

PR #116, opened 2025-07-21
Add `rnmi` interrupt mode

futog:
A new interrupt attribute mode has been added: `rnmi`.

The `rnmi` mode (resumable non-maskable interrupt) can be used with the `Smrnmi` extension.

Spec:
https://github.com/riscv/riscv-isa-manual/blob/20250508/src/rnmi.adoc

### DIFF
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -183,7 +183,7 @@ This attribute is incompatible with the `interrupt` attribute.
 NOTE: Be aware that compilers might have further restrictions on naked
 functions. Please consult your compiler's manual for more information.
 
-=== `+__attribute__((interrupt))+`, `+__attribute__((interrupt("supervisor")))+`, `+__attribute__((interrupt("machine")))+`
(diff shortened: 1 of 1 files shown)


futog:
> Do you know if anyone is working on a GCC patch?

I don't know, I couldn't find anything regarding this.

lenary (approved):
LGTM

topperc (approved):
LGTM

futog:
Hi @cmuellner, should I contact anybody else regarding this? What is the blocker of merging?

cmuellner:
> Given there is now a GCC patch and broad agreement here, I'm of the opinion that the LLVM patch is now unblocked. Does that seem reasonable?

The missing GCC patch was not blocking this PR. It was the missing approval from a GCC RISC-V maintainer. The GCC patch was done to avoid technical debt.

> it looks like GCC still has support for user interrupts, which were removed from this doc and removed from LLVM.

Thanks for reporting this! I've sent a patch.

PR #185, opened 2026-04-17
Propose extension bitmask assignments for newer standard RISC-V extensions

heturing:
Proposing bitmask assignments for some standard RISC-V extensions that currently do not have entries in the table.

I am working on LLVM support for newer RISC-V extensions already exposed by Linux `hwprobe`. These extensions currently do not have assignments in the bitmask assignment table, so I am proposing assignments for them here.

An open [PR](https://github.com/riscv-non-isa/riscv-c-api-doc/pull/171) already reserves (1,13), so this proposal starts from the next available slot and assigns entries sequentially:

- supm -> (1, 14)
- zicntr -> (1, 15)
- zihpm -> (1, 16)
- zfbfmin -> (1, 17)
- zvfbfmin -> (1, 18)
- zvfbfwma -> (1, 19)
- zicbom -> (1, 20)
- zaamo -> (1, 21)
- zalrsc -> (1, 22)
- zabha -> (1, 23)
- zalasr -> (1, 24)
- zicbop -> (1, 25)
- zicfilp -> (1, 26)

PR #187, opened 2026-05-17
Define extension bit for zicfiss

cmuellner:
Add zicfiss to the standard extension bitmask table.

Use group 1 bit 27 so the assignment does not conflict with the pending group 1 bits 14 through 26 proposed by riscv-c-api-doc PR #185.

### DIFF
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -1074,6 +1074,7 @@ Each queryable extension must have an associated `groupid` and `bitmask` that in
 | zalasr | 1 | 24
 | zicbop | 1 | 25
 | zicfilp | 1 | 26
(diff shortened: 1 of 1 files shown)


cmuellner:
GCC patch on list: https://gcc.gnu.org/pipermail/gcc-patches/2026-May/716957.html

kito-cheng (approved):
LGTM :)

lenary (approved):
LGTM. LLVM PR here: https://github.com/llvm/llvm-project/pull/201699

cmuellner:
Updated PR:
* rebased (#185 landed)
* updated commit message (no gap for #185 left)

PR #190, opened 2026-06-04
Add Zkr extension bitmask.

pz9115:
Add Zkr extension bitmask.

### DIFF
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -1029,6 +1029,7 @@ Each queryable extension must have an associated `groupid` and `bitmask` that in
 | zknh | 0 | 43
 | zksed | 0 | 44
 | zksh | 0 | 45
+| zkr | 1 | 28
 | zkt | 0 | 46
 | ztso | 0 | 47
 | zvbb | 0 | 48


src/c-api.adoc
```diff
@@ -1029,6 +1029,7 @@ Each queryable extension must have an associated `groupid` and `bitmask` that in
 | zknh | 0 | 43
 | zksed | 0 | 44
 | zksh | 0 | 45
+| zkr | 0 | 14
```
lenary: The bitmasks from (0, 0) to (0, 25) are reserved for single-letter extensions.

(1, 14) is already assigned by https://github.com/riscv-non-isa/riscv-c-api-doc/pull/185

I think the next free one is now (1, 28), given the other open PRs.

lenary (changes requested):

src/c-api.adoc
```diff
@@ -1029,6 +1029,7 @@ Each queryable extension must have an associated `groupid` and `bitmask` that in
 | zknh | 0 | 43
 | zksed | 0 | 44
 | zksh | 0 | 45
+| zkr | 0 | 28
```
lenary:
```suggestion
| zkr | 1 | 28
```

src/c-api.adoc
pz9115: Thank you for pointing this out, update the bitmask into (1,28).

src/c-api.adoc
pz9115: Fixed.

lenary:
Was this just missed when the other Crypto extensions were added?

Looking through, the extension just adds a seed CSR for crypto, but there is a note in the ISA manual that:

> Access to the entropy source (Zkr) in a system is usually carefully controlled. While the design supports unprivileged access to the entropy source, this is unlikely to be commonly used in an application processor, and so Zkr was not added as a profile option. This also means the roll-up Zk was not added as a profile option.

In particular, this bitmap aims to only represent things relevant to userspace extensions. I can see arguments for implementing this and for not doing so, I'd just like a little more motivation than a one line change :)

pz9115:
> Was this just missed when the other Crypto extensions were added?
Actually, I don't know the detail, maybe people can help give more information about this.

> 
> Looking through, the extension just adds a seed CSR for crypto, but there is a note in the ISA manual that:
> 
> > Access to the entropy source (Zkr) in a system is usually carefully controlled. While the design supports unprivileged access to the entropy source, this is unlikely to be commonly used in an application processor, and so Zkr was not added as a profile option. This also means the roll-up Zk was not added as a profile option.
> 
> In particular, this bitmap aims to only represent things relevant to userspace extensions. I can see arguments for implementing this and for not doing so, I'd just like a little more motivation than a one line change :)
OK, Let's wait for more information.

PR #192, opened 2026-07-13
Add function state attributes

kito-cheng:
Specify `__riscv_in`, `__riscv_out`, `__riscv_inout`, `__riscv_preserves` and 
`__riscv_new` keyword attributes, which describe the contract between a
function and its callers for architectural state such as the Xsfmm
matrix state.

Although the only state defined so far is the SiFive Xsfmm matrix
state, the attributes are not tied to it: the goal is that future
standard matrix extensions such as AME and VME, as well as vendor
extensions that add other architectural state, can reuse them by
defining new state strings.


lenary:
The clang implementation is making `__riscv_in(state)` and `__riscv_out(state)` mutually exclusive. Could we make that equivalent to `__riscv_inout(state)`?

PR #193, opened 2026-07-13
Add the riscv_rvv_vector_bits type attribute

kito-cheng:
Document the `riscv_rvv_vector_bits(N)` type attribute, which creates a fixed-length variant of a sizeless RVV data or mask type so that the type has a compile-time-known size while still working with the RVV intrinsics.

Add a new "Type Attributes" chapter that specifies:

- The GNU and `[[riscv::rvv_vector_bits]]` spellings;
- Why the attribute is needed instead of a GNU vector type.
- The applicable types (a single, non-segment data vector or a `vbool*_t` mask) and the `__riscv_v_fixed_vlen` macro that the argument is derived from.
- The N argument, that is `__riscv_v_fixed_vlen * LMUL` for data types and `__riscv_v_fixed_vlen / RATIO` for mask types, with a mismatch being a diagnosed error rather than undefined behavior.
- Size and layout, conversions, and the GNU vector operators, including the subscript operator whose out-of-range access is undefined behavior.
- The calling convention: Use standard fixed-length vector calling convention variant.

Also add the `__riscv_v_fixed_vlen` preprocessor macro, which expands to the fixed VLEN selected at compile time and is the basis for the attribute.

### DIFF
--- a/src/c-api.adoc
+++ b/src/c-api.adoc
@@ -46,7 +46,7 @@ The value of `+__riscv_v_min_vlen+` is defined by the following rules:
 is present.
 
 If multiple rules apply, the maximum value is taken.
-If none of the rules apply, `+__riscv_v_min_vlen+` is undefined.
+If none of the rules apply, `+__riscv_v_min_vlen+` is not defined.
 
 Examples:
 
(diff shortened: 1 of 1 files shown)


src/c-api.adoc
```diff
@@ -87,6 +87,37 @@ The value of `+__riscv_v_elen_fp+` is defined by the following rules:
...
+minimal VLEN mandated by the `-march` string, that is the same value as
+`+__riscv_v_min_vlen+`.
+
+NOTE: When `+__riscv_v_fixed_vlen+` is defined, the translation unit is built for
+this single fixed VLEN and may not run correctly on a processor whose VLEN differs
+from this value. For example, a translation unit built with
+`+__riscv_v_fixed_vlen+` equal to 128 may not run correctly on a processor with a
+VLEN of 256.
+
+`+__riscv_v_fixed_vlen+` is undefined when no fixed VLEN is selected (the
```
topperc, line 109: `undefined` -> `not defined` seems better to me, but I see we use `undefined` for the same thing elsewhere in this document. To me "undefined" could be read as it has an unknown value.

kito-cheng: Updated :)

PR #194, opened 2026-07-14
Use "not defined" instead of "undefined"

topperc:
"undefined" could be read as having an unknown or random value. "not defined" is more consistent with `#ifndef`


topperc:
CC @lenary @kito-cheng @cmuellner

lenary (approved):
LGTM

kito-cheng (approved):
LGTM

